PR #7, opened 2024-08-26
feat: first version of OmniRelayer

### LINKED ISSUE #5: [OmniBridge]: implement OmniRelayer for NEP141-connector
- [x] Track Near events
- [x] Track EVM events
- [x] Sign payload
- [x] Finalise transfer
- [x] Claim fee
- [x] Calculate fee profitability 
- [x] Integrate Wormhole as message passing

frolvanya:
NEAR -> ETH
- [x] Tracking NEAR events and filtering out events with `ft_on_transfer` call
  - [x] Filter out `sign_transfer_callback` calls as well to retreive logs
- [x] Sign transaction on relayer's behalf
- [x] Call `finalize_deposit_omni` method using `bridge-sdk`

ETH -> NEAR
- [x] Track `withdraw` events on eth blockchain
  - Side note: current rpc provider doesn't support `eth_newFilter` method, therefore we can't watch for incomming logs. As far as I understood, it can be fixed by switching to infura rpc provider
- [x] Call `finalize_withdraw` method using `bridge-sdk`

Other
- [x] Claim fee (I can't debug it for now, because `omni_deposit` doesn't exist yet)
- [x] Check if given fee is greater than expected fee (I've done simple implementation, but it needs to be modified later)
  - [x] Calculate fee on sender chain in native tokens
  - [x] Calculate fee on receiver chain in native tokens
  - [x] Using current exchange rate compare fee in USDT
  Possible problems:
  - Some tokens don't have a price (e.g coingecko is giving "0" as a price for those tokens)
  - Some tokens have a "fake" price (you can't actually exchange them using given price)
  Possible solutions:
  - To have a whitelist of "good" tokens that can be transferred using omni-bridge
- [x] Wormhole integration  
  - [x] for claiming fee
  - [x] for finalizing withdrawal
  
closes #5


omni-relayer/src/utils/eth.rs
```diff
@@ -0,0 +1,41 @@
...
+            return;
+        };
+        let Some(log_index) = decoded_log.log_index else {
+            log::warn!("No log index in log: {:?}", log);
+            return;
+        };
+
+        match connector
+            .finalize_withdraw(
+                primitive_types::H256::from_slice(tx_hash.as_slice()),
```
frolvanya, line 32: Maybe it's better to send tx_hash as a string to `bridge-sdk` and parse it to `H256` there. I don't think that introducing a whole crate for one type is a good idea

omni-relayer/src/defaults.rs
```diff
@@ -0,0 +1,16 @@
...
+pub const BRIDGE_TOKEN_FACTORY_ADDRESS_MAINNET: &str = "0x252e87862A3A720287E7fd527cE6e8d0738427A2";
+
+/// Testnet
+pub const NEAR_RPC_TESTNET: &str = "https://rpc.testnet.near.org/";
+pub const ETH_RPC_TESTNET: &str = "https://ethereum-sepolia.blockpi.network/v1/rpc/public";
+pub const ETH_CHAIN_ID_TESTNET: u64 = 11_155_111;
+pub const TOKEN_LOCKER_ID_TESTNET: &str = "omni-locker.test1-dev.testnet";
+pub const BRIDGE_TOKEN_FACTORY_ADDRESS_TESTNET: &str = "0xCF6462b9fcE5aF3E6c660c83453ECa18Ff468773";
+pub const NEAR_LIGHT_CLIENT_ETH_ADDRESS_TESTNET: &str =
+    "0x202cdf10bfa45a3d2190901373edd864f071d707";
```
karim-en, line 16: Let's move these consts to config file

omni-relayer/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+mod utils;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    pretty_env_logger::init();
+
+    let near_signer = startup::near::create_signer()?;
+    let connector = Arc::new(startup::build_connector(&near_signer)?);
+
+    tokio::spawn(startup::near::start_indexer(near_signer, connector.clone()));
```
karim-en, line 16: Let's add more async workers.
1- near_sing_transfer_worker
2- eth_fin_transfer_worker
3- near_claim_fee_worker
4- eth_track_events_worker
5- near_track_events_order

The events, transaction hashes, and nonces should be stored in the Redis DB.
Also, additional workers could be added for transaction status tracking to avoid blocking the primary workers.

omni-relayer/src/startup/eth.rs
```diff
@@ -0,0 +1,48 @@
...
+            .context("Failed to parse ETH rpc provider as url")?,
+    );
+
+    let ws_provider = ProviderBuilder::new()
+        .on_ws(WsConnect::new(defaults::ETH_WS_MAINNET))
+        .await
+        .context("Failed to initialize WS provider")?;
+
+    let latest_block = http_provider.get_block_number().await?;
+    let from_block = latest_block.saturating_sub(1000);
```
karim-en, line 29: Let's store the last processed block in the Redis db

karim-en (changes requested):

omni-relayer/src/main.rs
frolvanya: > Let's add more async workers.
> 1- near_sing_transfer_worker
> 2- eth_fin_transfer_worker
> 3- near_claim_fee_worker
> 4- eth_track_events_worker
> 5- near_track_events_order

Can you please take a look at the new version? The first thought idea that came to my mind was to use simple channels, but:
a) it made code more complicated without any real reason to do so
b) Now as we use channels we can't really call two methods at the same time

I guess you meant something different, can you please elaborate a bit then? For example, in previous approach we had `sign_transfer` async call directly from `near_track_events`. If we want separate these workers I don't see other approach than using channels, but as I've mentioned above it only made things worse in my opinion

omni-relayer/src/defaults.rs
frolvanya: Did you mean to create a new module like `config.rs` and move these constants there or you wanted to created a real config file like `config.toml`?

omni-relayer/src/startup/eth.rs
frolvanya: It's easy to do with near lake indexer, but I believe that we can't track each block on eth blockchain. We only are able to handle blocks that are already filtered out (aka only blocks with events on our bridge account). Is it good enough or it's better to dig a bit and try to update db on each block?

omni-relayer/src/utils/eth.rs
frolvanya: > send tx_hash as a string to `bridge-sdk` and parse it to `H256` there

@kiseln What do you think about this? Can we do it like that or it would introduce a lot of mess on sdk's side?

omni-relayer/src/startup/eth.rs
frolvanya: Instead of tweaking eth event tracker, we can modify "continue from last processed block" algorithm. I think there's no point to start indexer from block that it's older than X days (I guess X should be 3 or even 1), because person most likely wouldn't expect relayer to help him with finalizing transactions after a day of sending it. So we probably can update last processed block in redis on every upcomming event on bridge token factory and the finished algo will be something like:
```pseudocode
if final_block - redis.get("eth_last_processed_block") > avg_blocks_produced_per_day * X then
  start indexer from final_block - 10_000
else
  start indexer from redis.get("eth_last_processed_block")
```

frolvanya:
@karim-en How much data should be stored in redis db? Should we set an expiration time for each log (e.g 1 day) or configure a simple log rotation (e.g keep only last 1000 logs)?

omni-relayer/src/defaults.rs
karim-en: a full config file like config.toml or config.json

omni-relayer/src/startup/eth.rs
karim-en: I think  `redis.get("eth_last_processed_block").unwrap_or(config.eth_start_from_block)` should be enough

omni-relayer/src/main.rs
karim-en: Please take a look at this implementation https://github.com/Near-One/fast-bridge-service/blob/master/src/main.rs.
The workers are synced just by reading from the redis-db, but we optimize it with observer pattern.
This is needed to guarantee data and process integrity; the service should correctly resume if it was killed or restarted.

frolvanya: The code is far from being the cleanest, but the core logic works, and everything is synchronized through Redis. Could you please take a look to see if this is the desired implementation?

omni-relayer/src/defaults.rs
frolvanya: Now it feels a bit odd to have both defaults.rs and config.toml. Should I move everything to config.toml?

omni-relayer/src/utils/eth.rs
kiseln: I'm not exactly sure. I assume it's a light weight crate and I'd rather keep strong typing. Is it possible to re-export these types from the sdk so the client doesn't have to specify the dependency?

frolvanya: > Is it possible to re-export these types from the sdk so the client doesn't have to specify the dependency?

Not sure, but I'll check this. Then for this PR I'll just leave this crate

omni-relayer/src/workers/near.rs
```diff
@@ -0,0 +1,255 @@
...
+    redis_client: redis::Client,
+    connector: Arc<Nep141Connector>,
+) -> Result<()> {
+    let redis_connection = redis_client.get_multiplexed_tokio_connection().await?;
+
+    loop {
+        let mut redis_connection_clone = redis_connection.clone();
+        let Some(mut events) = utils::redis::get_events(
+            &mut redis_connection_clone,
+            config.redis.eth_deposit_events.clone(),
```
olga24912, line 170: These don't have to be Ethereum events. In the future, they could be events from Solana, for example. In this file, interactions with blockchains other than NEAR should be encapsulated. Maybe this function should be moved to eth.rs.

omni-relayer/src/workers/near.rs
```diff
@@ -0,0 +1,255 @@
...
+                        };
+                        let Some(log_index) = deposit_log.log_index else {
+                            log::warn!("No log index in log: {:?}", deposit_log);
+                            return;
+                        };
+
+                        match eth_proof::get_proof_for_event(
+                            H256::from_slice(tx_hash.as_slice()),
+                            log_index,
+                            &config.mainnet.eth_rpc_http_url,
```
olga24912, line 207: Sometimes it is hardcoded mainnet, sometimes it is hardcoded testnet. Sounds strange.

omni-relayer/src/workers/eth.rs
```diff
@@ -0,0 +1,83 @@
...
+            if let Ok(withdraw_log) =
+                serde_json::from_str::<Log<crate::startup::eth::Withdraw>>(&event)
+            {
+                handlers.push(tokio::spawn({
+                    let config = config.clone();
+                    let mut redis_connection = redis_connection.clone();
+                    let connector = connector.clone();
+
+                    async move {
+                        log::info!("Decoded log: {:?}", withdraw_log);
```
olga24912, line 45: We also should check fee here.

omni-relayer/src/workers/near.rs
```diff
@@ -0,0 +1,255 @@
...
+                        else {
+                            warn!("Expected InitTransferEvent, got: {:?}", event);
+                            return;
+                        };
+
+                        info!(
+                            "Received InitTransferEvent: {}",
+                            transfer_message.origin_nonce.0
+                        );
+
```
olga24912, line 59: we also should check fee here.

omni-relayer/src/workers/near.rs
```diff
@@ -0,0 +1,255 @@
...
+                        match connector
+                            .sign_transfer(
+                                transfer_message.origin_nonce.into(),
+                                Some(config.testnet.token_locker_id),
+                                transfer_message.fee.into(),
+                            )
```
olga24912, line 65: We probably want to provide our NEAR account here.

omni-relayer/src/utils/redis.rs
```diff
@@ -0,0 +1,75 @@
...
+        warn!("Failed to serialize event: {:?}", event);
+        return;
+    };
+
+    if let Err(err) = redis_connection
+        .hset_nx::<&str, F, String, ()>(key, field, serialized_event)
+        .await
+    {
+        warn!("Failed to add event to redis db: {}", err);
+    }
```
olga24912, line 62: In case of error we should retry to add event in the future.

omni-relayer/src/startup/eth.rs
```diff
@@ -0,0 +1,120 @@
...
+    let ws_provider = ProviderBuilder::new()
+        .on_ws(WsConnect::new(config.mainnet.eth_rpc_ws_url.clone()))
+        .await
+        .context("Failed to initialize WS provider")?;
+
+    let latest_block = http_provider.get_block_number().await?;
+    let from_block =
+        utils::redis::get_last_processed_block(&mut redis_connection, "eth_last_processed_block")
+            .await
+            .map_or_else(|| latest_block.saturating_sub(10_000), |block| block);
```
olga24912, line 53: 10_000 -- should be added to config

omni-relayer/src/startup/near.rs
```diff
@@ -0,0 +1,90 @@
...
+pub fn create_signer() -> Result<InMemorySigner> {
+    info!("Creating NEAR signer");
+
+    let account_id = std::env::var("NEAR_ACCOUNT_ID")
+        .context("Failed to get `NEAR_ACCOUNT_ID` env variable")?
+        .parse()?;
+
+    let private_key = std::env::var("NEAR_PRIVATE_KEY")
+        .context("Failed to get `NEAR_PRIVATE_KEY` env variable")?
+        .parse()?;
```
olga24912, line 20: Add the ability to pass the path to near_credential and extract the key from it.

omni-relayer/src/utils/near.rs
```diff
@@ -0,0 +1,117 @@
...
+            } => {
+                utils::redis::add_event(
+                    redis_connection,
+                    &config.redis.near_sign_transfer_events,
+                    message_payload.nonce.0.to_string(),
+                    log,
+                )
+                .await;
+            }
+            Nep141LockerEvent::FinTransferEvent { .. }
```
olga24912, line 90: This event should be handled the same way as InitTransferEvent in some cases.

omni-relayer/src/utils/near.rs
```diff
@@ -0,0 +1,117 @@
...
+        .cloned()
+        .collect()
+}
+
+fn is_nep_locker_event(config: &config::Config, receipt: &ReceiptView) -> bool {
+    receipt.receiver_id == config.testnet.token_locker_id
+        && matches!(
+            receipt.receipt,
+            ReceiptEnumView::Action { ref actions, .. } if actions.iter().any(|action| {
+                matches!(action, ActionView::FunctionCall { method_name, .. } if method_name == "ft_on_transfer" || method_name == "sign_transfer_callback")
```
olga24912, line 114: "update_transfer_fee", "fin_transfer_callback". 

Should we event filter by method_name?

olga24912 (changes requested):

omni-relayer/config.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[mainnet]
+eth_rpc_http_url = "https://eth.llamarpc.com"
+eth_rpc_ws_url = "wss://eth-mainnet.g.alchemy.com/v2/API-KEY"
+bridge_token_factory_address = "0x252e87862A3A720287E7fd527cE6e8d0738427A2"
+
+[testnet]
+near_rpc_url = "https://rpc.testnet.near.org/"
+eth_rpc_url = "https://ethereum-sepolia.blockpi.network/v1/rpc/public"
+eth_chain_id = 11_155_111
```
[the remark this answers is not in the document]
frolvanya: It makes sense. But for now it's easier for me to use current implementation as I'm using testnet for near and mainnet for eth, so I'll change it a bit later

omni-relayer/src/workers/near.rs
frolvanya: yep, this issue will go away as soon as I change config structure as you advised before

omni-relayer/src/workers/near.rs
frolvanya: Hmm, I don't think so. I'm checking it right before adding this event to redis in utils/near.rs. The time difference between adding event to db and processing it in worker is at most 10 seconds, so I doubt that we need to check it here again. However, maybe it makes more sense to move my initial check from utils to this worker

omni-relayer/src/workers/near.rs
frolvanya: We're providing it in startup/mod.rs build_connector function and `sign_transfer` just uses provided token_locker_id that is provided in connector

omni-relayer/src/utils/near.rs
frolvanya: > "update_transfer_fee", "fin_transfer_callback"

But relayer has nothing to do with these calls yet. `update_transfer_fee` doesn't require any action from relayer, as far as I know

omni-relayer/src/workers/near.rs
frolvanya: For now I renamed it to "finalized_transfers", but in the future I'd need to pass simple `Deposit` structure instead of `Log<customDeposit>`. I can't do it for now, because it expects only `eth` proof for now. I need to think how to refactor it. Right now I'm thinking about storing directly `ClaimFeeArgs` in redis instead of event structure

omni-relayer/src/utils/near.rs
olga24912: require. You should check fee one more time and handle transaction. In second case it can be transfer from Eth->Solana for example.

omni-relayer/src/workers/near.rs
frolvanya: Done. It was fixed by storing `FinTransferArgs` and `ClaimTransferArgs` in db, instead of storing custom eth log structure. However, I'm not sure if we should move `claim_fee` method in eth.rs worker, claim_fee feels more attached to token locker account and near blockchain itself, but let me know if you still think that it'd be better to move `claim_fee` to eth.rs

omni-relayer/src/workers/eth.rs
frolvanya: I believe that we only need to check/claim fee during `NEAR -> X` transfers, not vice versa. At least that's what I remember from our discussion with @karim-en

omni-relayer/src/workers/near.rs
frolvanya: @olga24912 Let me know where it would be better to move the claim fee check in your opinion. Right now it's being called from utils/near.rs right after we receive log. So there're 3 options here:
1) Leave check in utils module
2) Move check to near worker
3) Have two checks: in utils and in near worker

omni-relayer/src/utils/near.rs
frolvanya: InitTransfer should trigger `sign_transfer` method, but why do we need to sign anything after the transfer being finalized? In future we can add `claim_fee` that would claim fee on the sender's chain, but it has nothing to do with signing something

I don't see any extra steps from relayer after getting a FinTransferEvent in this graph as well: https://docs.nearone.org/doc/omni-bridge-IBwRPSc6SG#h-eth-→-near-transfer. We just need to catch `Withdraw` event and call `fin_transfer` right after

omni-relayer/src/workers/near.rs
olga24912: This field is missing: https://github.com/Near-One/rainbow-bridge-sdk-rs/blob/omni-token-support/bridge-sdk/connectors/omni-connector/src/omni_connector.rs#L324

olga24912 (changes requested):

omni-relayer/src/startup/near.rs
```diff
@@ -0,0 +1,104 @@
...
+use near_crypto::InMemorySigner;
+use near_jsonrpc_client::JsonRpcClient;
+use near_lake_framework::{LakeConfig, LakeConfigBuilder};
+
+use crate::{config, utils};
+
+pub fn create_signer(file: Option<String>) -> Result<InMemorySigner> {
+    info!("Creating NEAR signer");
+
+    let account_id = std::env::var("NEAR_ACCOUNT_ID")
```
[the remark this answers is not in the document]
frolvanya: Here's my near-credentials file. How can we get `account_id` from it? I mean we technically can get it from filename (just strip `.json` extension, but it's not very reliable)

`{"seed_phrase_hd_path":"...","master_seed_phrase":"...","implicit_account_id":"...","public_key":"...","private_key":"..."}`

omni-relayer/src/utils/fee.rs
```diff
@@ -0,0 +1,78 @@
...
+    let sender_fee = match sender {
+        OmniAddress::Near(_) => 0.03 * get_price_by_symbol("near").await?,
+        OmniAddress::Eth(_) => 0.00005 * get_price_by_symbol("ethereum").await?,
+        OmniAddress::Sol(_) => 0.001 * get_price_by_symbol("solana").await?
+    };
+    let recipient_fee = match recipient {
+        OmniAddress::Near(_) => 0.03 * get_price_by_symbol("near").await?,
+        OmniAddress::Eth(_) => 0.00005 * get_price_by_symbol("ethereum").await?,
+        OmniAddress::Sol(_) => 0.001 * get_price_by_symbol("solana").await?
+    };
```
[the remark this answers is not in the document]
frolvanya: Yep, this is just a hardcoded part. Haven't finished it yet

omni-relayer/config.toml
```diff
@@ -0,0 +1,13 @@
+[redis]
+url = "redis://127.0.0.1/"
+
+[near]
+rpc_url = "https://rpc.testnet.near.org/"
+token_locker_id = "omni-locker.test1-dev.testnet"
+
+[eth]
+rpc_http_url = "https://eth.llamarpc.com"
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/7/commits/9f683d7143b57ad5f4954a35bb85d493bf27d33b

omni-relayer/config.toml
```diff
@@ -0,0 +1,13 @@
...
+rpc_ws_url = "wss://eth-mainnet.g.alchemy.com/v2/API-KEY"
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/7/commits/9f683d7143b57ad5f4954a35bb85d493bf27d33b

omni-relayer/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+
+const CONFIG_FILE: &str = "config.toml";
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    pretty_env_logger::init();
+
+    let mut config = toml::from_str::<config::Config>(&std::fs::read_to_string(CONFIG_FILE)?)?;
+
+    config.inject_api_keys()?;
```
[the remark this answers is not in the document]
frolvanya: We decided that putting api key directly in config file is much more convenient, so this can be ignored now

omni-relayer/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use anyhow::Result;
+use log::{error, info};
+
+mod config;
+mod startup;
+mod utils;
+mod workers;
+
+const CONFIG_FILE: &str = "config.toml";
```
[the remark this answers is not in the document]
frolvanya: Added config argument in https://github.com/Near-One/omni-bridge/pull/7/commits/9fa3451d547a858d3409fdab0b397e62074d787c

omni-relayer/src/startup/near.rs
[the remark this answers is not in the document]
frolvanya: Used filename to retrieve account_id https://github.com/Near-One/omni-bridge/pull/7/commits/3ec5a7803af43e7688ac4d137e52c4c1a1ee2876

Current logic: try to get some info from file -> if not succeeded, then try to get data from env variables -> if not succeeded, then panic

omni-relayer/src/startup/near.rs
```diff
@@ -0,0 +1,104 @@
...
+    .await
+    {
+        Some(block_height) => block_height,
+        None => utils::near::get_final_block(jsonrpc_client).await?,
+    };
+
+    info!("NEAR Lake will start from block: {}", start_block_height);
+
+    LakeConfigBuilder::default()
+        .testnet()
```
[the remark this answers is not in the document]
frolvanya: Added `network` field to config in https://github.com/Near-One/omni-bridge/pull/7/commits/fe757aab0a8f043eb7bca7cbb01b53af408674e2

omni-relayer/src/utils/near.rs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                }
+
+                utils::redis::add_event(
+                    redis_connection,
+                    utils::redis::NEAR_INIT_TRANSFER_EVENTS,
+                    transfer_message.origin_nonce.0.to_string(),
+                    log,
+                )
+                .await;
```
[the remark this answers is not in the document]
frolvanya: Mostly it's for debugging purposes. I'm planning to introduce a new key in redis db (e.g `not_enough_fee`) and check this table once in a while and recalculate fee/gas

omni-relayer/src/workers/near.rs
frolvanya: Sorry, I still don't understand why it's missing. We are setting near signer here: https://github.com/Near-One/omni-bridge/blob/ed02e1475c33aab169d644523ed7a2fa5e4e4950/omni-relayer/src/startup/mod.rs#L28-L29

and then creating new `InMemorySigner` here:
https://github.com/Near-One/rainbow-bridge-sdk-rs/blob/omni-token-support/bridge-sdk/connectors/omni-connector/src/omni_connector.rs#L481-L495

Moreover, we have the same logic in `sign_transfer` and it works perfectly fine as we tested it today:
https://github.com/Near-One/rainbow-bridge-sdk-rs/blob/omni-token-support/bridge-sdk/connectors/omni-connector/src/omni_connector.rs#L341-L372

karim-en (approved):

frolvanya:
@olga24912 Can you approve this PR, please? I'll apply all changes from your comments that are left in a new branch

olga24912 (approved):

PR #16, opened 2024-09-05
Add EVM contracts for Omni Bridge

kiseln:
This is a modified version of the contracts from https://github.com/Near-One/rainbow-token-connector

The primary difference is using chain signatures instead of light client proofs to mint tokens


evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -0,0 +1,318 @@
...
+        if (ECDSA.recover(hashed, signatureData) != nearBridgeDerivedAddress) {
+            revert InvalidSignature();
+        }
+
+        require(_isBridgeToken[_nearToEthToken[bridgeDeposit.token]], "ERR_NOT_BRIDGE_TOKEN");
+        BridgeToken(_nearToEthToken[bridgeDeposit.token]).mint(bridgeDeposit.recipient, bridgeDeposit.amount);
+
+        completedTransfers[bridgeDeposit.nonce] = true;
+
+        emit Deposit(bridgeDeposit.token, bridgeDeposit.amount, bridgeDeposit.recipient);
```
karim-en, line 202: We need to emit the nonce to claim the fee on Near side.
Also, we need to pass a `string claim_recipient/fee_recipient` to identify the fee recipient

evm/bridge-token-factory/scripts/deploy-factory.js
```diff
@@ -0,0 +1,53 @@
+require("dotenv").config();
+const { ethers, upgrades } = require("hardhat");
+
+const BRIDGE_TOKEN_IMPL_ADDRESS = process.env.BRIDGE_TOKEN_IMPL_ADDRESS;
```
karim-en, line 4: I don't like this script format. We should use hardhat tasks and config files, as we did for other projects.

karim-en (changes requested):

evm/bridge-token-factory/contracts/Borsh.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+library Borsh {
```
karim-en, line 4: We must figure out how to share this `Brosh.sol` and `SelectivePausableUpgradable` with other contracts.
So we have two options here: put all the contracts under the same project but in different directories like it is done in `openzeppelin-contracts` repo; Or move the shared code to a separate project.

kiseln: Makes sense but we can think about it when we add another project

evm/bridge-token-factory/scripts/deploy-factory.js
kiseln: Done 7ae04b99707283238932465ab7bb228efbb57cb0

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
kiseln: Added it to signature as discussed 616d939c21e3ae02cec87bc2b27caa61a68335f5

karim-en (changes requested):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -0,0 +1,330 @@
...
+        string recipient,
+        address indexed tokenEthAddress
+    );
+
+    event Deposit(
+        string indexed token,
+        uint256 amount,
+        address recipient,
+        uint128 indexed nonce,
+        string indexed feeRecipient
```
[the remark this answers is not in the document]
kiseln: I indexed `feeRecipient` because an off-chain service may want to search for events related to a specific recipient (to collect all the fees, for example). Yes, it will have to hash the account Id before performing the search.

`token` can be useful for analytics. `recipient`, I agree, is the least useful

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -0,0 +1,330 @@
...
+        CheckAccountAndToken
+    }
+
+    // We removed ProofConsumer from the list of parent contracts and added this gap
+    // to preserve storage layout when upgrading to the new contract version.
+    uint256[54] private __gap;
+
+    mapping(address => string) private _ethToNearToken;
+    mapping(string => address) private _nearToEthToken;
+    mapping(address => bool) private _isBridgeToken;
```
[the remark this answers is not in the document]
kiseln: It's been there for a while. Looks redundant because we can use the other two mappings to validate if token has been added

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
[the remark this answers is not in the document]
kiseln: Removed indexing of `feeRecipient` as hashing would complicate the logic when claiming fees on Near.

Still not sure about `token` and `recipient`. I view events not only from POV of utility that allows us to execute some logic on Near, but also as a source of additional information about user activity. This way it makes sense to keep these properties.

karim-en (approved):

PR #18, opened 2024-09-06
feat: added necessary types for `bridge-sdk`

frolvanya:


near/omni-types/src/lib.rs
```diff
@@ -243,3 +243,9 @@ pub type Nonce = u128;
 pub fn stringify<T: std::fmt::Display>(item: T) -> String {
     item.to_string()
 }
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct TransferLog {
+    pub message_payload: TransferMessagePayload,
```
karim-en, line 249: Please use types from the near_event.
If you want it as struct instead of enum, then can refactor the near_events with serd flatten macro

karim-en (changes requested):

karim-en (approved):

PR #21, opened 2024-09-10
Omni Bridge EVM factory with wormhole

kiseln:
Added a derived contract that implements omni bridge factory with wormhole as a message passing mechanism


evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
```diff
@@ -0,0 +1,51 @@
...
+        address wormholeAddress,
+        uint8 consistencyLevel
+    ) external initializer {
+        initialize(_tokenImplementationAddress, _nearBridgeDerivedAddress);
+        _wormhole = IWormhole(wormholeAddress);
+        _consistencyLevel = consistencyLevel;
+    }
+
+    function depositExtension(BridgeDeposit memory bridgeDeposit) internal override {
+        _wormhole.publishMessage(
```
karim-en, line 32: this method is payable 
https://github.com/wormhole-foundation/wormhole/blob/a3284d66d5da7c0480262d5f7c93d7cc8322f6a1/ethereum/contracts/bridge/Bridge.sol#L358

karim-en (changes requested):

karim-en:
@kiseln kind reminder

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
```diff
@@ -0,0 +1,51 @@
...
+    function publishMessage(
+        uint32 nonce,
+        bytes memory payload,
+        uint8 consistencyLevel
+    ) external payable returns (uint64 sequence);
+}
+
+contract BridgeTokenFactoryWormhole is BridgeTokenFactory {
+    IWormhole private _wormhole;
+    uint8 private _consistencyLevel;
```
olga24912, line 17: What about make it const or add some recommendation for this value: https://wormhole.com/docs/build/reference/consistency-levels/

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
```diff
@@ -0,0 +1,51 @@
...
+    ) external initializer {
+        initialize(_tokenImplementationAddress, _nearBridgeDerivedAddress);
+        _wormhole = IWormhole(wormholeAddress);
+        _consistencyLevel = consistencyLevel;
+    }
+
+    function depositExtension(BridgeDeposit memory bridgeDeposit) internal override {
+        _wormhole.publishMessage(
+            wormholeNonce,
+            abi.encode(bridgeDeposit.token, bridgeDeposit.amount, bridgeDeposit.feeRecipient, bridgeDeposit.nonce),
```
olga24912, line 34: Maybe use `encodeWithSignature` for encoding and specify what kind of event it is, so that it's easier to decode using the prefix.

evm/bridge-token-factory/contracts/test/TestWormhole.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+contract TestWormhole {
```
olga24912, line 4: https://github.com/wormhole-foundation/wormhole/blob/a3284d66d5da7c0480262d5f7c93d7cc8322f6a1/ethereum/contracts/Implementation.sol#L15

Maybe use an implementation closer to the real one. It seems to be fairly simple.

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
olga24912: https://github.com/Near-One/omni-bridge/blob/main/near/omni-types/src/prover_result.rs#L15

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
```diff
@@ -0,0 +1,51 @@
...
+        );
+
+        wormholeNonce++;
+
+    }
+
+    function withdrawExtension(string memory token, uint128 amount, string memory recipient) internal override {
+        _wormhole.publishMessage(
+            wormholeNonce,
+            abi.encode(token, amount, recipient),
```
olga24912, line 45: https://github.com/Near-One/omni-bridge/blob/main/near/omni-types/src/prover_result.rs#L9

olga24912: https://github.com/Near-One/omni-bridge/blob/main/near/omni-types/src/lib.rs#L197

We should also add info about fee and sender

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
```diff
@@ -0,0 +1,51 @@
...
+
+interface IWormhole {
+    function publishMessage(
+        uint32 nonce,
+        bytes memory payload,
+        uint8 consistencyLevel
+    ) external payable returns (uint64 sequence);
+}
+
+contract BridgeTokenFactoryWormhole is BridgeTokenFactory {
```
olga24912, line 15: And also we should generate event on NewBridgeToken: https://github.com/Near-One/omni-bridge/blob/main/near/omni-types/src/prover_result.rs#L23

olga24912 (changes requested):

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: This is chain-specific so has to be parametrized

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: Thanks, added a39d2b1af1d04b5493db58b8a6299d8cd900c0aa

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: I added sender but fee is not currently implemented for withdrawals (it's not exclusive to wormhole). Let's discuss and add in the future PRs

evm/bridge-token-factory/contracts/test/TestWormhole.sol
kiseln: I'd rather keep it simple. We just want to test that we pass correct parameters and we can do it with this simple setup

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: Not a bad idea, @karim-en ?
It can be useful for example if we read all events from Wormhole API so we don't know the type of the events

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: f781851ad9ce65355f64c75da1ef143f5338411d

evm/bridge-token-factory/test/BridgeToken.js
```diff
@@ -36,10 +36,12 @@ describe('BridgeToken', () => {
     const bridgeToken = await BridgeTokenInstance.deploy()
     await bridgeToken.waitForDeployment()
 
+    const nearBridgeDeriveAddress = "0xa966f32b64caaee9211d674e698cb72100b5e792";
```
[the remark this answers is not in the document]
kiseln: I'm not sure there is a library. There is a file in examples repo that does the calculations https://github.com/near-examples/near-multichain/blob/main/src/services/kdf.js

Anyway, I think it's not necessary. We hardcode signatures signed by this address, so we can also hardcode the address itself. It should have no affect on the integrity of the tests.

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
karim-en: yeh, let's add uint8 at the beginning

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
karim-en: This should be covered in this issue 
https://github.com/Near-One/omni-bridge/issues/24

evm/bridge-token-factory/test/BridgeToken.js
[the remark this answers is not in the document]
kiseln: Even if it changes we don't need to change the tests. Or if we change the tests we will still have to hardcode signatures. But I guess it can be useful to include for the deploy scripts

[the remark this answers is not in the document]
kiseln: 30524e8d7a9bbe3429981fc21cf13e40a55ca7e6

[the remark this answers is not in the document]
karim-en: Thanks

evm/bridge-token-factory/contracts/BridgeTokenFactoryWormhole.sol
kiseln: Added ee552d5

olga24912 (approved):

karim-en (approved):

PR #23, opened 2024-09-12
NEP141-Locker: implement storage deposit

karim-en:
- Make the storage upgradable by introducing `TransferMessageStorage` type.
- Replace `unwrap_or_else` with `SdkExpect`
- Introduce storage management.
- Add API to retrieve required storage deposits for different actions.

### DIFF
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -144,6 +144,7 @@ impl FungibleTokenReceiver for Contract {
             sender: OmniAddress::Near(sender_id.to_string()),
         };
 
+        // TODO: add native token as fee attachment
         let required_storage_balance =
             self.add_transfer_message(self.current_nonce, &transfer_message, &sender_id);
         self.update_storage_balance(
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -1,4 +1,4 @@
-use errors::{SdkExpect, SdkUnwrap};
+use errors::SdkExpect;
 use near_plugins::{
     access_control, access_control_any, pause, AccessControlRole, AccessControllable, Pausable,
     Upgradable,
@@ -140,7 +140,7 @@ impl FungibleTokenReceiver for Contract {
             origin_nonce: U128(self.current_nonce),
             token: env::predecessor_account_id(),
             amount,
-            recipient: msg.parse().sdk_unwrap(),
+            recipient: msg.parse().sdk_expect("ERR_PARSE_MSG"),
             fee: U128(0), // TODO get fee from msg
             sender: OmniAddress::Near(sender_id.to_string()),
         };
@@ -614,8 +614,7 @@ impl Contract {
         message_owner: &AccountId,
     ) -> NearToken {
         let storage_usage = env::storage_usage();
-        self.insert_raw_transfer(nonce, transfer_message, message_owner)
-            .sdk_unwrap();
+        self.insert_raw_transfer(nonce, transfer_message, message_owner);
         env::storage_byte_cost().saturating_mul((env::storage_usage() - storage_usage).into())
     }
 
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -638,7 +638,7 @@ impl Contract {
             .get(&owner)
             .sdk_expect("ERR_ACCOUNT_NOT_REGISTERED");
 
-        storage.available = storage.available.saturating_sub(refund);
+        storage.available = storage.available.saturating_add(refund);
         self.accounts_balances.insert(&owner, &storage);
 
         transfer_message
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -51,7 +51,7 @@ enum StorageKey {
     Factories,
     FinalisedTransfers,
     TokensMapping,
-    Accounts,
+    AccountsBalances,
 }
 
 #[derive(AccessControlRole, Deserialize, Serialize, Copy, Clone)]
@@ -122,8 +122,7 @@ pub struct Contract {
     pub tokens_to_address_mapping: LookupMap<(ChainKind, AccountId), OmniAddress>,
     pub mpc_signer: AccountId,
     pub current_nonce: Nonce,
-    pub accounts: LookupMap<AccountId, StorageBalance>,
-    pub account_storage_usage: u64,
+    pub accounts_balances: LookupMap<AccountId, StorageBalance>,
 }
 
 #[near]
@@ -170,13 +169,11 @@ impl Contract {
             tokens_to_address_mapping: LookupMap::new(StorageKey::TokensMapping),
             mpc_signer,
             current_nonce: nonce.0,
-            accounts: LookupMap::new(StorageKey::Accounts),
-            account_storage_usage: 0,
+            accounts_balances: LookupMap::new(StorageKey::AccountsBalances),
         };
 
         contract.acl_init_super_admin(near_sdk::env::predecessor_account_id());
         contract.acl_grant_role("DAO".to_owned(), near_sdk::env::predecessor_account_id());
-        contract.measure_account_storage_usage();
         contract
     }
 
@@ -597,20 +594,6 @@ impl Contract {
         }
     }
 
-    fn measure_account_storage_usage(&mut self) {
-        let initial_storage_usage = env::storage_usage();
-        let tmp_account_id = "a".repeat(64).parse().unwrap();
-        self.accounts.insert(
-            &tmp_account_id,
-            &StorageBalance {
-                total: NearToken::from_yoctonear(0),
-                available: NearToken::from_yoctonear(0),
-            },
-        );
-        self.account_storage_usage = env::storage_usage() - initial_storage_usage;
-        self.accounts.remove(&tmp_account_id);
-    }
-
     fn insert_raw_transfer(
         &mut self,
         nonce: u128,
@@ -637,7 +620,6 @@ impl Contract {
     }
 
     fn add_fin_transfer(&mut self, chain_kind: ChainKind, nonce: u128) -> NearToken {
-        // TODO: the storage size can be calculated on init
         let storage_usage = env::storage_usage();
         require!(
             self.finalised_transfers.insert(&(chain_kind, nonce)),
@@ -652,11 +634,11 @@ impl Contract {
         required_balance: NearToken,
         attached_deposit: NearToken,
     ) {
-        let refund = match self.accounts.get(&account_id) {
+        let refund = match self.accounts_balances.get(&account_id) {
             Some(mut storage) => {
                 if storage.available >= required_balance {
                     storage.available = storage.available.saturating_sub(required_balance);
-                    self.accounts.insert(&account_id, &storage);
+                    self.accounts_balances.insert(&account_id, &storage);
                     Some(attached_deposit)
                 } else {
                     None
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -24,7 +24,7 @@ use omni_types::{
     ChainKind, MetadataPayload, NearRecipient, Nonce, OmniAddress, SignRequest, TransferMessage,
     TransferMessagePayload, UpdateFee,
 };
-use storage::TransferMessageStorage;
+use storage::{TransferMessageStorage, TransferMessageStorageValue};
 
 mod errors;
 mod storage;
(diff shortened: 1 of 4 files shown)


near/nep141-locker/src/lib.rs
```diff
@@ -112,11 +118,12 @@ pub trait Prover {
 pub struct Contract {
     pub prover_account: AccountId,
     pub factories: LookupMap<ChainKind, OmniAddress>,
-    pub pending_transfers: LookupMap<Nonce, TransferMessage>,
+    pub pending_transfers: LookupMap<Nonce, TransferMessageStorage>,
     pub finalised_transfers: LookupSet<(ChainKind, Nonce)>,
     pub tokens_to_address_mapping: LookupMap<(ChainKind, AccountId), OmniAddress>,
     pub mpc_signer: AccountId,
     pub current_nonce: Nonce,
+    pub accounts_balances: LookupMap<AccountId, StorageBalance>,
```
kiseln, line 126: This storage management mechanism adds quite a bit of complexity. Is this a normal practice for smart contracts on Near? I was thinking if required payments are small enough we can just collect them without refunding

karim-en: Please check these links
https://docs.near.org/concepts/storage/storage-staking
https://nomicon.io/Standards/StorageManagement
https://docs.near.org/tutorials/fts/registering-accounts

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    pub fn required_balance_for_fin_transfer(&self) -> NearToken {
+        let key_len = borsh::to_vec(&(ChainKind::Eth, 0_u128))
+            .sdk_expect("ERR_BORSH")
+            .len() as u64;
+
+        env::storage_byte_cost().saturating_mul((Self::get_basic_storage() + key_len).into())
+    }
+
+    fn get_basic_storage() -> u64 {
```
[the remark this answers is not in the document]
karim-en: The `EXTRA_BYTES_RECORD` is from here: https://github.com/near/nearcore/blob/cdb8055d917180146722b2ca0e6314c8b48eebf8/core/parameters/res/runtime_configs/parameters.yaml#L30C6-L31C1
The 1 byte is for the enum prefix that we use.

near/nep141-locker/src/lib.rs
karim-en: Also, take into account that these balances will be used not just to pay for storage but also to attach fee in native/base token

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        env::storage_byte_cost()
+            .saturating_mul((Self::get_basic_storage() + key_len + value_len).into())
+    }
+
+    pub fn required_balance_for_init_transfer(
+        &self,
+        recipient: OmniAddress,
+        sender: OmniAddress,
+    ) -> NearToken {
```
karim-en, line 108: This is the most controversial part; we currently can't mock those values like for account id because:
1- we don't limit the String for OmniAddress::Near & OmniAddress::Sol
2- the OmniAddress::Near recipient could be with message `account:msg`

near/nep141-locker/src/storage.rs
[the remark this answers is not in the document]
kiseln: Sorry, it's not clear which enum are we talking about and what the first value is for

near/nep141-locker/src/lib.rs
kiseln: If refund values are negligible, would it be reasonable to remove this logic to reduce contract complexity?

near/nep141-locker/src/storage.rs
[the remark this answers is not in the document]
karim-en: This [one](https://github.com/Near-One/omni-bridge/blob/78122c47dcb29357aec2aeac9dc390930f2b7225/near/nep141-locker/src/lib.rs#L50), it is the storage key prefix for a specific map.
Check [this](https://docs.near.org/build/smart-contracts/anatomy/collections#error-prone-patterns)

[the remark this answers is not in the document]
karim-en: PS @kiseln, all the collections under the hood call this function to write to the storage.
https://docs.rs/near-sdk/latest/near_sdk/env/fn.storage_write.html

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
+use near_contract_standards::storage_management::{StorageBalance, StorageBalanceBounds};
+use near_sdk::{assert_one_yocto, borsh};
+use near_sdk::{env, near_bindgen, AccountId, NearToken};
+
+use crate::*;
+
+pub type TransferMessageStorageValue = (TransferMessage, AccountId);
```
olga24912, line 7: Maybe it's better to add named fields. Right now, it's not clear what AccountId means.

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+            }
             PromiseResult::Failed => Err(PromiseError::Failed),
         }
     }
+
+    fn insert_raw_transfer(
+        &mut self,
+        nonce: u128,
+        transfer_message: &TransferMessage,
+        sender_id: &AccountId,
```
olga24912, line 601:
```suggestion
        message_owner: &AccountId,
```

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+    ) -> Option<Vec<u8>> {
```
olga24912, line 602: We never use the return value. Should we return this?

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+
+    fn add_transfer_message(
+        &mut self,
+        nonce: u128,
+        transfer_message: &TransferMessage,
+        message_owner: &AccountId,
+    ) -> NearToken {
+        let storage_usage = env::storage_usage();
+        self.insert_raw_transfer(nonce, transfer_message, message_owner);
+        env::storage_byte_cost().saturating_mul((env::storage_usage() - storage_usage).into())
```
olga24912, line 618: Should we use flush here? I remember we had similar issues somewhere before.

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+        }
+        .unwrap_or_else(|| {
+            attached_deposit
+                .checked_sub(required_balance)
+                .sdk_expect("Not enough storage deposited")
+        });
+
+        if !refund.is_zero() {
+            Promise::new(account_id).transfer(refund);
+        }
```
olga24912, line 678: And why is the logic like this? In my mind, I imagined subtracting from the attached funds first and then covering the remaining amount from the storage balance if needed.

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+    fn remove_transfer_message(&mut self, nonce: u128) -> TransferMessage {
+        let storage_usage = env::storage_usage();
+        let (transfer_message, owner) = self
+            .pending_transfers
+            .remove(&nonce)
+            .map(|m| m.into_main())
+            .sdk_expect("ERR_TRANSFER_NOT_EXIST");
+
+        let refund =
+            env::storage_byte_cost().saturating_mul((storage_usage - env::storage_usage()).into());
```
olga24912, line 630: flush?

near/nep141-locker/src/lib.rs
```diff
@@ -549,9 +587,94 @@ impl Contract {
...
+
+        let mut storage = self
+            .accounts_balances
+            .get(&owner)
+            .sdk_expect("ERR_ACCOUNT_NOT_REGISTERED");
+
+        storage.available = storage.available.saturating_sub(refund);
```
olga24912, line 637: add?

olga24912 (changes requested):

near/nep141-locker/src/lib.rs
karim-en: No, because we are using `near-sdk::collections`

near/nep141-locker/src/lib.rs
karim-en: we are using `near-sdk::collections` not `near-sdk::store`

near/nep141-locker/src/lib.rs
karim-en: 28d0e76fa508dc51d355728f5a9ea42425802e91

near/nep141-locker/src/lib.rs
karim-en: 8e64e520b784216e4f62693aa872d294c7b1d3f9

near/nep141-locker/src/lib.rs
karim-en: c17a99899605f38d0a64b652822281e475d53587

near/nep141-locker/src/lib.rs
karim-en: Like this deab232ed0f853ca79d8ffbdd82924ccb83681c3?

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+        assert_one_yocto();
+        let account_id = env::predecessor_account_id();
+        let mut storage = self
+            .storage_balance_of(&account_id)
+            .sdk_expect("The account is not registered");
+        let to_withdraw = amount.unwrap_or(storage.available);
+        storage.total = storage
+            .total
+            .checked_sub(to_withdraw)
+            .sdk_expect("The amount is greater than the total storage balance");
```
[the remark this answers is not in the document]
karim-en: 7d9a6e7b5221a5a503aa288e1aa33388b14c0309

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+        let key_len = borsh::to_vec(&(ChainKind::Eth, 0_u128))
+            .sdk_expect("ERR_BORSH")
+            .len() as u64;
+
+        env::storage_byte_cost().saturating_mul((Self::get_basic_storage() + key_len).into())
+    }
+
+    fn get_basic_storage() -> u64 {
+        const EXTRA_BYTES_RECORD: u64 = 40;
+        const EXTRA_KEY_PREFIX_LEN: u64 = 1;
```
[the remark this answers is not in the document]
karim-en: check this conversation 
https://github.com/Near-One/omni-bridge/pull/23#discussion_r1765333937

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+        enum RefTransferMessageStorage<'a> {
+            V0((&'a TransferMessage, &'a AccountId)),
+        }
+
+        borsh::to_vec(&RefTransferMessageStorage::V0((message, account)))
+    }
+}
+
+#[near_bindgen]
+impl Contract {
```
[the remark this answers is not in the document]
karim-en: Not relevant to this PR.
Anyway, the contract code is fully isolated, so I don't see much need for naming it.

near/nep141-locker/src/storage.rs
karim-en: The goal was to avoid extra clones, but yeh, it's not worth it

near/nep141-locker/src/storage.rs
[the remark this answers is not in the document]
karim-en: Check this conversation 
https://github.com/Near-One/omni-bridge/pull/23#discussion_r1765333937

near/nep141-locker/src/storage.rs
```diff
@@ -0,0 +1,142 @@
...
+            available: NearToken::from_yoctonear(0),
+        })
+        .sdk_expect("ERR_BORSH")
+        .len() as u64;
+
+        env::storage_byte_cost()
+            .saturating_mul((Self::get_basic_storage() + key_len + value_len).into())
+    }
+
+    pub fn required_balance_for_init_transfer(
```
olga24912, line 130: We don't use this function. Should we remove this?

near/nep141-locker/src/storage.rs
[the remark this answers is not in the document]
olga24912: Maybe just remove this logic and do it like in other places: measure storage before and after, and calculate the difference?

olga24912 (changes requested):

olga24912 (approved):

PR #33, opened 2024-09-23
NEP141-Locker: parse fee from msg

karim-en:


near/omni-types/src/lib.rs
```diff
@@ -193,6 +193,12 @@ impl fmt::Display for NearRecipient {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct InitTransferMsg {
+    pub recip: OmniAddress,
```
kiseln, line 198: Why not `recipient`?

olga24912 (approved):

near/omni-types/src/lib.rs
karim-en: The idea was to reduce the message size so the transaction could be signed using the ledger.
Anyway, let me rename it to `recipient` and return it to `recip` if it is needed

karim-en: 209b11bca92e4f4b4a4ba7b9a1a6f619745ac6aa

kiseln (approved):

near/omni-types/src/lib.rs
kiseln: New ledger screen is pretty big it seems

karim-en: @kiseln it isn't about the screen, but about the payload limits for a near wallet on ledger

PR #37, opened 2024-09-24
 EVM-Prover: move rainbow bridge proof verification code

karim-en:
This PR aims to eliminate the need for an EVM-Prover proxy, which should reduce verification latency and gas consumption. Also, the number of maintained contracts should be reduced.
The source contract is https://github.com/Near-One/rainbow-bridge/blob/master/contracts/near/eth-prover/src/lib.rs


kiseln:
Just two comments regarding bringing the EVM types in.

1. Do we want to rewrite the types to use alloy primitives? (not now necessarily but in general)
2. Do we want to keep these types in `omni-types` or is it better to encapsulate them in the prover crate? I am not sure they should be used by any other component of the OmniBridge

karim-en:
> Just two comments regarding bringing the EVM types in.
> 
> 1. Do we want to rewrite the types to use alloy primitives? (not now necessarily but in general)
> 2. Do we want to keep these types in `omni-types` or is it better to encapsulate them in the prover crate? I am not sure they should be used by any other component of the OmniBridge

1. probably yeh, I want to do that, but we need more time to rewrite it and implement tests. Spicily the rlp decoding.
2. Good question. I worry about two things: contract binary size & conflict dependency issues, and until now, I haven't noticed any of these issues. Because the compiler is good at optimizing unlinked code.
So, I propose restructuring it when we have at least one of these issues.

https://rust-unofficial.github.io/patterns/patterns/structural/small-crates.html

near/omni-types/src/prover_args.rs
```diff
@@ -15,6 +15,8 @@ pub struct VerifyProofArgs {
 pub struct EvmVerifyProofArgs {
     pub proof_kind: ProofKind,
     pub proof: EvmProof,
+    pub min_header_height: Option<u64>,
+    pub max_header_height: Option<u64>,
```
karim-en, line 19: It doesn't make sense to pass these arguments by the sender; the connector contract should pass them, so we need to move them somewhere upper or remove them and relay on the nonce instead.

kiseln (approved):

near/omni-types/src/prover_args.rs
karim-en: Just removed it b8e29324908172ad23993bb641b0acb214c69434

kiseln (approved):

olga24912:
Do I understand correctly that we will still be supporting the old prover?

The old prover has quite a lot of functionality that is not presented in this contract, and it is still widely used.

karim-en:
> Do I understand correctly that we will still be supporting the old prover?
> 
> The old prover has quite a lot of functionality that is not presented in this contract, and it is still widely used.

Not sure that it's really used by someone else, but yeh, we will continue support it for some period of time.

near/omni-prover/evm-prover/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+/// None if the block number is not part of the canonical chain yet.
+#[ext_contract(evm_client)]
+pub trait EvmClient {
+    #[result_serializer(borsh)]
+    fn block_hash_safe(&self, #[serializer(borsh)] index: u64) -> Option<H256>;
+}
+
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct EvmProver {
```
olga24912, line 28: Do we want to add roles? The ability to pause the contract? The ability to change the light_client?

Or are we planning to simply deploy a new contract and change the prover in the higher-level contract if necessary?

near/omni-prover/evm-prover/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// @param proof contains relevant information to verify data is valid
+    ///
+    /// Patricia Trie: https://eth.wiki/en/fundamentals/patricia-tree
+    /// Patricia Img:  https://ethereum.stackexchange.com/questions/268/ethereum-block-architecture/6413#6413
+    ///
+    /// Verification:  https://github.com/slockit/in3/wiki/Ethereum-Verification-and-MerkleProof#receipt-proof
+    /// Article:       https://medium.com/@ouvrard.pierre.alain/merkle-proof-verification-for-ethereum-patricia-tree-48f29658eec
+    /// Python impl:   https://gist.github.com/mfornet/0ff283274c0162f1cca45966bccf69ee
+    ///
+    fn verify_trie_proof(expected_root: H256, key: Vec<u8>, proof: &Vec<Vec<u8>>) -> Vec<u8> {
```
olga24912, line 130: We must support the old prover as well. This function is an exact copy-paste. In the short term, it’s easier to do it this way, and it also avoids unnecessary dependencies. However, it might make sense in the future to move this code to a separate library."

near/omni-types/src/evm/header.rs
```diff
@@ -0,0 +1,159 @@
+pub use ethereum_types::{Address, Bloom, H256, H64, U256, U64};
+use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+
+use super::utils::keccak256;
+
+#[derive(Default, Debug, Clone)]
```
olga24912, line 6: BorshSerialize, BorshDeserialize?

near/omni-types/src/evm/header.rs
```diff
@@ -0,0 +1,159 @@
...
+pub struct BlockHeader {
```
olga24912, line 7: It would be great to have one place for all standard types at some point.

olga24912 (approved):

near/omni-prover/evm-prover/src/lib.rs
karim-en: we planning to simply deploy a new contract and change the prover in the higher-level contract

near/omni-prover/evm-prover/src/lib.rs
karim-en: Yeh, agree

near/omni-types/src/evm/header.rs
karim-en: Unfortunately, the types don't support borsh, and I don't want to move all the `eth-types`, also, as mentioned by @kiseln, we are going to replace the types to alloy at some point.

PR #41, opened 2024-09-26
Add OmniBridge chain id to properly validate omnichain recipient address

kiseln:
Currently tokens are only minted for Ethereum signature because we hardcoded 0 as OmniAddress enum variant

### DIFF
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -36,6 +36,7 @@ contract BridgeTokenFactory is
 
     address public tokenImplementationAddress;
     address public nearBridgeDerivedAddress;
+    uint8 private _omniBridgeChainId;
 
     mapping(uint128 => bool) public completedTransfers;
     uint128 public initTransferNonce; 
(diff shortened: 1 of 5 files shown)


evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -36,6 +36,7 @@ contract BridgeTokenFactory is
 
     address public tokenImplementationAddress;
     address public nearBridgeDerivedAddress;
+    uint8 private _omniBridgeChainId;
```
karim-en, line 39: Is there are any benefits of making it private? If not, then let's make it public

karim-en (changes requested):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
kiseln: Just couldn't see a reason anyone would need this and want to keep API smaller

kiseln: 4887048bd4a026929e22219c733a0e5921faf3cc

karim-en (approved):

PR #43, opened 2024-09-26
EVM-Prover: fix header rlp decoding

karim-en:


near/omni-types/src/evm/header.rs
```diff
@@ -55,7 +55,7 @@ impl<'a> CustomRlpIter<'a> {
             Ok(result) => {
                 self.consumed_bytes += result.as_raw().len();
                 self.index += 1;
-                result.as_val()
+                Ok(Some(result.as_val::<T>()?))
```
frolvanya, line 58: We don't need `::<T>`, it will guess type based on return type

```suggestion
                Ok(Some(result.as_val()?))
```

frolvanya (approved):

PR #44, opened 2024-09-26
feat: added serialization for `OmniAddress`

frolvanya:


kiseln (approved):

near/omni-types/src/lib.rs
```diff
@@ -37,7 +38,38 @@ impl FromStr for H160 {
...
+                        c
+                    }
+                }
+                _ => c,
+            };
+
+            result.push(c);
+        }
+
+        write!(f, "0x{}", result)
```
karim-en, line 72: I would like to avoid these kinds of operations on-chain, especially since using the `alloy_primitives::Keccak256` isn't acceptable; if it is needed just for the relayer, then please implement a separate method `to_eip_55_string` and use the hash function from utils

karim-en (changes requested):

near/omni-types/src/lib.rs
frolvanya: So should I just delete it?

karim-en: You can keep the `to_eip_55_checksum` as a public method if you want, but in the fmt & serd let's use `write!(f, "0x{}", hex::encode(self.0))`

frolvanya: Done https://github.com/Near-One/omni-bridge/pull/44/commits/8cedbc8a42fee426aad3b5fcd9530c07c408f659

karim-en (approved):

PR #45, opened 2024-09-29
feat: pay fee in native token

karim-en:


near/nep141-locker/src/lib.rs
```diff
@@ -398,7 +470,7 @@ impl Contract {
                 "STORAGE_ERR: The transfer recipient is omitted"
             );
 
-            let amount_to_transfer = U128(transfer_message.amount.0 - transfer_message.fee.0);
+            let amount_to_transfer = U128(transfer_message.amount.0 - transfer_message.fee.fee.0);
```
olga24912, line 473: Not consistent with part on Ethereum. In Ethereum amount not included fee.

near/nep141-locker/src/lib.rs
```diff
@@ -440,6 +516,7 @@ impl Contract {
 
             promise.into()
         } else {
+            required_balance = self.add_fin_transfer(&transfer_message.get_transfer_id(), &None);
```
olga24912, line 519: It looks like the required_balance is overwritten two lines later.

olga24912 (changes requested):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -248,18 +257,53 @@ contract BridgeTokenFactory is
 
         BridgeToken(tokenAddress).burn(msg.sender, amount + fee);
 
-        initTransferExtension(initTransferNonce, token, amount, fee, recipient, msg.sender);
+        uint256 extensionValue = msg.value - nativeFee;
```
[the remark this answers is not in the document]
karim-en: It will definitely panic 
https://docs.soliditylang.org/en/latest/080-breaking-changes.html
https://github.com/NomicFoundation/hardhat/issues/1209

near/nep141-locker/src/lib.rs
```diff
@@ -379,12 +446,17 @@ impl Contract {
...
-            transfer_message.get_origin_chain(),
-            transfer_message.origin_nonce.0,
-        );
+        let mut required_balance;
 
         if let OmniAddress::Near(recipient) = &transfer_message.recipient {
+            required_balance = self.add_fin_transfer(
+                &transfer_message.get_transfer_id(),
+                &Some(NativeFee {
+                    amount: transfer_message.amount,
```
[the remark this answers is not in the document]
karim-en: 395b2cf708c572e5e618f3374dc0ca33398466af

kiseln:
I think simplest would be to only keep native fee and remove token fee. Do we have requriements from Defuse on what they prefer?

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -248,18 +257,53 @@ contract BridgeTokenFactory is
 
         BridgeToken(tokenAddress).burn(msg.sender, amount + fee);
```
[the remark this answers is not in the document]
karim-en: 6e60c19210a9b8ca93adce648cb7c2bed4bd48fd

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -61,13 +62,20 @@ contract BridgeTokenFactory is
         uint8 decimals;
     }
 
+    struct ClaimFeePayload {
```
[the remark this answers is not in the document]
karim-en: I think we can use enum for payloads to fix this problem

near/nep141-locker/src/lib.rs
```diff
@@ -273,22 +281,76 @@ impl Contract {
         }
     }
 
+    #[payable]
+    pub fn sign_claim_native_fee(&mut self, claim_payload: ClaimNativeFeePayload) -> Promise {
```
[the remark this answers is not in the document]
karim-en: 4f5ac0b40a2234377b9ae68aab1b8f7da2d81563

near/omni-types/src/near_events.rs
```diff
@@ -25,6 +25,10 @@ pub enum Nep141LockerEvent {
         signature: SignatureResponse,
         metadata_payload: MetadataPayload,
     },
+    SingCalimNativeFeeEvent {
```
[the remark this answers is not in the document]
karim-en: c1feddcae016f28f24db9ae39534fabb5a616e55

karim-en:
> I think simplest would be to only keep native fee and remove token fee. Do we have requriements from Defuse on what they prefer?

Yeh, I also prefer paying the fees in a native token, but we have already implemented paying fee in the transferred token, so I don't think that it is a good idea to remove it.

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -39,6 +39,7 @@ contract BridgeTokenFactory is
     uint8 public omniBridgeChainId;
 
     mapping(uint128 => bool) public completedTransfers;
+    mapping(uint128 => bool) public claimedFee;
```
karim-en, line 42: As I see, we can remove `claimedFee` and just use the `completedTransfers`.
@kiseln @olga24912 do you agree?

near/nep141-locker/src/lib.rs
```diff
@@ -259,6 +266,7 @@ impl Contract {
                     "Only sender can update fee"
                 );
 
+                // TODO: attach deposit
```
karim-en, line 269: The new fee can't be lower than the previous fee.

near/omni-types/src/near_events.rs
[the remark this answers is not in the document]
kiseln: Sing -> Sign :)

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: Actually, no, this will not work because we can't use the near nonce instead of the origin nonce, so this map is needed.

olga24912: No, we can't

near/nep141-locker/src/lib.rs
```diff
@@ -479,30 +579,56 @@ impl Contract {
...
                 == Some(fin_transfer.emitter_address),
-            "Unknown factory"
+            "ERR_UNKNOWN_FACTORY"
         );
 
         let message = self.remove_transfer_message(fin_transfer.nonce.0);
         let fee = message.amount.0 - fin_transfer.amount.0;
 
+        if message.get_origin_chain() == ChainKind::Near && message.fee.native_fee.0 != 0 {
+            Promise::new(fin_transfer.fee_recipient.clone())
```
olga24912, line 614:
```suggestion
            Promise::new(native_fee_recipient)
```

olga24912 (changes requested):

near/nep141-locker/src/lib.rs
karim-en: @olga24912 please verify these changes 
1c5f596ea14994c783c27109c89dda5d139e7cd2

near/nep141-locker/src/lib.rs
karim-en: @olga24912 please check this one

PR #64, opened 2024-10-08
feat: unwrap wnear if that's the incoming token

kiseln:


kiseln:
https://github.com/Near-One/omni-bridge/issues/60

near/nep141-locker/src/lib.rs
```diff
@@ -493,15 +501,24 @@ impl Contract {
...
             let amount_to_transfer = U128(transfer_message.amount.0 - transfer_message.fee.fee.0);
-            let mut promise = match recipient.message {
-                Some(message) => ext_token::ext(transfer_message.token.clone())
-                    .with_static_gas(FT_TRANSFER_CALL_GAS)
-                    .with_attached_deposit(ONE_YOCTO)
-                    .ft_transfer_call(recipient.target, amount_to_transfer, None, message),
-                None => ext_token::ext(transfer_message.token.clone())
-                    .with_static_gas(FT_TRANSFER_GAS)
+
+            let mut promise = if transfer_message.token == self.wnear_account_id {
```
karim-en, line 505: We shouldn't unwrap the token if it is a `ft_transfer_call` (with msg).
FYI: in the current eNear connector, we are wrapping the native near to be able to do a ft_transfer_call

karim-en (changes requested):

near/nep141-locker/src/lib.rs
kiseln: Updated dfdf9a65bb26269ce21b5832a4d2d1ce19878e6d

olga24912 (approved):

PR #68, opened 2024-10-09
[OmniBridge]: Fix bind token in NEP141-Locker

olga24912:
* Check the deposit and refund
* ix the annotation of bind_token_callback to ensure consistency with the prover
* Make bind_token_callback payable 
* https://github.com/orgs/Near-One/projects/3/views/1?pane=issue&itemId=82324067&issue=Near-One%7Comni-bridge%7C61

### DIFF
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -652,25 +652,42 @@ impl Contract {
                 Self::ext(env::current_account_id())
                     .with_attached_deposit(env::attached_deposit())
                     .with_static_gas(BIND_TOKEN_CALLBACK_GAS)
-                    .bind_token_callback(),
+                    .bind_token_callback(near_sdk::env::predecessor_account_id()),
             )
     }
 
     #[private]
+    #[payable]
     pub fn bind_token_callback(
         &mut self,
+        predecessor_account_id: AccountId,
         #[callback_result]
-        #[serializer(borsh)]
         call_result: Result<ProverResult, PromiseError>,
     ) {
         let Ok(ProverResult::DeployToken(deploy_token)) = call_result else {
-            env::panic_str("Invalid proof message")
+            Self::refund(predecessor_account_id, env::attached_deposit());
+            env::log_str("ERROR: Invalid proof message");
+            return;
         };
+
+        let storage_usage = env::storage_usage();
 
         self.tokens_to_address_mapping.insert(
             &(deploy_token.token_address.get_chain(), deploy_token.token),
             &deploy_token.token_address,
         );
+
+        let required_deposit = env::storage_byte_cost()
+            .saturating_mul((env::storage_usage().saturating_sub(storage_usage)).into());
+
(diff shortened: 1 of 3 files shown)


frolvanya (approved):

near/nep141-locker/src/lib.rs
```diff
@@ -689,25 +689,41 @@ impl Contract {
...
         self.tokens_to_address_mapping.insert(
             &(deploy_token.token_address.get_chain(), deploy_token.token),
             &deploy_token.token_address,
         );
+
+        let required_deposit = env::storage_byte_cost()
+            .saturating_mul((env::storage_usage().saturating_sub(storage_usage)).into());
+
+        if env::attached_deposit() < required_deposit {
+            Self::refund(predecessor_account_id, env::attached_deposit());
```
karim-en, line 720: Logically, it is correct, but three calls for a refund in one function is too much.
Please add `required_balance_for_bind_token`, and remove the refund logic.
Even if the refund was needed in this scenario, then it should be wrapped into closure, function, or additional callback. 

https://stackoverflow.com/questions/66629383/is-there-a-way-to-return-from-a-block-expression

eg:
```
    #[private]
    #[payable]
    pub fn bind_token_callback(
        &mut self,
        predecessor_account_id: AccountId,
        #[callback_result] call_result: Result<ProverResult, PromiseError>,
    ) {
        let (refund_amount, err) = (|| {
            let Ok(ProverResult::DeployToken(deploy_token)) = call_result else {
                return (env::attached_deposit(), "ERROR: Invalid proof message");
            };

            let storage_usage = env::storage_usage();

            self.tokens_to_address_mapping.insert(
                &(deploy_token.token_address.get_chain(), deploy_token.token),
                &deploy_token.token_address,
            );

            let required_deposit = env::storage_byte_cost()
                .saturating_mul((env::storage_usage().saturating_sub(storage_usage)).into());

            if env::attached_deposit() < required_deposit {
                return (
                    env::attached_deposit(),
                    "ERROR: The deposit is not sufficient to cover the storage.",
                );
            }

            let refund_amount = env::attached_deposit().saturating_sub(required_deposit);

            return (refund_amount, "");
        })();

        env::log_str(err);
        Self::refund(predecessor_account_id, refund_amount);
    }
```

karim-en (changes requested):

near/nep141-locker/src/lib.rs
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/68/commits/ce4544731e0c199be224ef32d89bc67f696a9e19

I decided to implement two callbacks after all, in order to fully roll back a failed transaction. This is because, in the case of insufficient deposit, we weren't reverting the changes in the LookUp Map.

olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/68/commits/6f70a0574f47f7aa0e02fcb5b6aaf399f88694ab

karim-en (approved):

PR #70, opened 2024-10-11
feat: claim native fee

### LINKED ISSUE #57: [OmniBridge]: initiate signClaimNativeFee and claimNativeFee in the Relayer

frolvanya:
closes #57,#58

### DIFF
--- a/omni-relayer/src/config.rs
+++ b/omni-relayer/src/config.rs
@@ -1,5 +1,6 @@
 use alloy::primitives::Address;
 use near_primitives::types::AccountId;
+use omni_types::OmniAddress;
 
 #[derive(Debug, Clone, serde::Deserialize)]
 pub struct Config {
@@ -35,4 +36,5 @@ pub struct Evm {
     pub chain_id: u64,
     pub bridge_token_factory_address: Address,
     pub block_processing_batch_size: u64,
+    pub relayer: OmniAddress,
 }
--- a/omni-relayer/src/config.rs
+++ b/omni-relayer/src/config.rs
@@ -27,6 +27,7 @@ pub struct Near {
     pub rpc_url: String,
     pub token_locker_id: AccountId,
     pub credentials_path: Option<String>,
+    pub relayer_address_on_evm: OmniAddress,
 }
 
 #[derive(Debug, Clone, serde::Deserialize)]
@@ -36,5 +37,4 @@ pub struct Evm {
     pub chain_id: u64,
     pub bridge_token_factory_address: Address,
     pub block_processing_batch_size: u64,
-    pub relayer: OmniAddress,
 }
--- a/omni-relayer/src/config.rs
+++ b/omni-relayer/src/config.rs
@@ -1,6 +1,25 @@
 use alloy::primitives::Address;
-use near_primitives::types::AccountId;
-use omni_types::OmniAddress;
+use alloy::signers::k256::ecdsa::SigningKey;
+use alloy::signers::local::LocalSigner;
+use near_primitives::{borsh::BorshDeserialize, types::AccountId};
+use omni_types::{OmniAddress, H160};
+
+fn derive_evm_address_from_private_key() -> OmniAddress {
+    let decoded_private_key = hex::decode(
+        std::env::var("ETH_PRIVATE_KEY").expect("Failed to get `ETH_PRIVATE_KEY` env variable"),
+    )
+    .expect("Failed to decode `ETH_PRIVATE_KEY`");
+
+    let secret_key = SigningKey::from_slice(&decoded_private_key)
+        .expect("Failed to create a `SecretKey` from the provided private key");
+
+    let signer = LocalSigner::from_signing_key(secret_key);
+
+    OmniAddress::Eth(
+        H160::try_from_slice(signer.address().as_slice())
+            .expect("Failed to create `OmniAddress` from the derived public key"),
+    )
+}
 
 #[derive(Debug, Clone, serde::Deserialize)]
 pub struct Config {
@@ -27,7 +46,6 @@ pub struct Near {
     pub rpc_url: String,
     pub token_locker_id: AccountId,
     pub credentials_path: Option<String>,
-    pub relayer_address_on_evm: OmniAddress,
 }
 
 #[derive(Debug, Clone, serde::Deserialize)]
@@ -37,4 +55,7 @@ pub struct Evm {
     pub chain_id: u64,
     pub bridge_token_factory_address: Address,
     pub block_processing_batch_size: u64,
+
+    #[serde(default = "derive_evm_address_from_private_key")]
+    pub relayer_address_on_eth: OmniAddress,
 }
--- a/omni-relayer/src/utils/near.rs
+++ b/omni-relayer/src/utils/near.rs
@@ -119,18 +119,20 @@ pub async fn handle_streamer_message(
             Nep141LockerEvent::SignClaimNativeFeeEvent {
                 ref claim_payload, ..
             } => {
-                utils::redis::add_event(
-                    redis_connection,
-                    utils::redis::NEAR_SIGN_CLAIM_NATIVE_FEE_EVENTS,
-                    claim_payload
-                        .nonces
-                        .iter()
-                        .map(|nonce| nonce.0.to_string())
-                        .collect::<Vec<_>>()
-                        .join(","),
-                    log,
-                )
-                .await;
+                if claim_payload.recipient == config.near.relayer_address_on_evm {
+                    utils::redis::add_event(
+                        redis_connection,
+                        utils::redis::NEAR_SIGN_CLAIM_NATIVE_FEE_EVENTS,
+                        claim_payload
+                            .nonces
+                            .iter()
+                            .map(|nonce| nonce.0.to_string())
+                            .collect::<Vec<_>>()
+                            .join(","),
+                        log,
+                    )
+                    .await;
+                }
             }
             Nep141LockerEvent::LogMetadataEvent { .. } => {}
         }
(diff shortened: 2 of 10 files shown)


omni-relayer/src/utils/near.rs
```diff
@@ -88,10 +88,51 @@ pub async fn handle_streamer_message(
...
+                }
+            }
+            Nep141LockerEvent::ClaimFeeEvent {
+                ref transfer_message,
+                ref native_fee_recipient,
+            } => {
+                if native_fee_recipient == &config.near.relayer_address_on_evm {
+                    utils::redis::add_event(
+                        redis_connection,
+                        utils::redis::NEAR_FIN_TRANSFER_EVENTS,
```
olga24912, line 112: I don't like that in the case of a ClaimFeeEvent, you're saving the event in NEAR_FIN_TRANSFER_EVENT. This is not accurate. I understand that it's done because the further processing is the same, but in that case, the column should be named differently.

For example, SIGN_CLAIM_NATIVE_FEE_QUEUE

omni-relayer/src/utils/near.rs
```diff
@@ -88,10 +88,51 @@ pub async fn handle_streamer_message(
...
+            } => {
+                if nonce.is_none() {
+                    utils::redis::add_event(
+                        redis_connection,
+                        utils::redis::NEAR_FIN_TRANSFER_EVENTS,
+                        transfer_message.origin_nonce.0.to_string(),
+                        log,
+                    )
+                    .await;
+                }
```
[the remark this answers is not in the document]
olga24912: I also think we should add to the FinTransferEvent native_fee_recipient, for not handle unnecessary events. @karim-en what do you think?

omni-relayer/src/utils/near.rs
```diff
@@ -88,10 +88,51 @@ pub async fn handle_streamer_message(
...
+            }
+            Nep141LockerEvent::SignClaimNativeFeeEvent {
+                ref claim_payload, ..
```
olga24912, line 120: Will this be parsed correctly?

https://github.com/Near-One/omni-bridge/blob/develop/near/nep141-locker/src/lib.rs#L364

omni-relayer/src/utils/near.rs
```diff
@@ -88,10 +88,51 @@ pub async fn handle_streamer_message(
...
+                    claim_payload
+                        .nonces
+                        .iter()
+                        .map(|nonce| nonce.0.to_string())
+                        .collect::<Vec<_>>()
+                        .join(","),
+                    log,
+                )
+                .await;
+            }
```
olga24912, line 134: Also, it would be useful to check the recipient to handle only our transactions.

omni-relayer/src/workers/evm.rs
```diff
@@ -73,3 +73,65 @@ pub async fn finalize_withdraw(
...
+    redis_client: redis::Client,
+    connector: Arc<OmniConnector>,
+) -> Result<()> {
+    let redis_connection = redis_client.get_multiplexed_tokio_connection().await?;
+
+    loop {
+        let mut redis_connection_clone = redis_connection.clone();
+        let Some(events) = utils::redis::get_events(
+            &mut redis_connection_clone,
+            utils::redis::ETH_WITHDRAW_EVENTS.to_string(),
```
olga24912: Why ETH_WITHDRAW_EVENTS?

omni-relayer/src/workers/evm.rs
```diff
@@ -73,3 +73,65 @@ pub async fn finalize_withdraw(
...
+                        let Nep141LockerEvent::SignClaimNativeFeeEvent { .. } = event else {
+                            warn!("Expected SignClaimNativeFeeEvent, got: {:?}", event);
+                            return;
+                        };
+
+                        log::info!("Received SignClaimNativeFeeEvent log");
+
+                        match connector.evm_claim_native_fee_with_log(event).await {
+                            Ok(tx_hash) => {
+                                log::info!("Finalized withdraw: {:?}", tx_hash);
```
olga24912, line 115: Finalized withdraw?

omni-relayer/src/workers/evm.rs
```diff
@@ -73,3 +73,65 @@ pub async fn finalize_withdraw(
...
+                                utils::redis::remove_event(
+                                    &mut redis_connection,
+                                    utils::redis::ETH_WITHDRAW_EVENTS,
```
olga24912, line 118: ETH_WITHDRAW_EVENTS?

omni-relayer/src/workers/near.rs
```diff
@@ -206,3 +206,77 @@ pub async fn claim_fee(redis_client: redis::Client, connector: Arc<OmniConnector
...
+                    let mut redis_connection = redis_connection.clone();
+                    let connector = connector.clone();
+
+                    async move {
+                        let Nep141LockerEvent::FinTransferEvent {
+                            ref transfer_message,
+                            ..
+                        } = event
+                        else {
+                            warn!("Expected FinTransferEvent, got: {:?}", event);
```
olga24912, line 246: Or ClaimFee event...

omni-relayer/src/workers/near.rs
```diff
@@ -206,3 +206,77 @@ pub async fn claim_fee(redis_client: redis::Client, connector: Arc<OmniConnector
...
+                            return;
+                        };
+
+                        log::info!("Received FinTransferEvent log");
```
olga24912, line 250: or ClaimFee event

omni-relayer/src/workers/near.rs
```diff
@@ -206,3 +206,77 @@ pub async fn claim_fee(redis_client: redis::Client, connector: Arc<OmniConnector
...
+
+                        match connector
+                            .sign_claim_native_fee(
+                                vec![transfer_message.origin_nonce.into()],
+                                config.near.relayer_address_on_evm,
+                            )
+                            .await
+                        {
+                            Ok(tx_hash) => {
+                                log::info!("Claimed native fee: {:?}", tx_hash);
```
olga24912, line 260: Signed

olga24912 (changes requested):

omni-relayer/src/utils/near.rs
frolvanya: What do you mean? I don't see a problem here. Since we don't care about signature I just retrieved the payload

omni-relayer/src/utils/near.rs
frolvanya: Fixed in fdf5afdbf258aad37a0caf64b758d9c403cf77dc

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in 0f3efd2

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in f218c04b15a5f0c5f72a4440b203d11449842b17

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in f218c04b15a5f0c5f72a4440b203d11449842b17

omni-relayer/src/workers/near.rs
frolvanya: Fixed in f218c04b15a5f0c5f72a4440b203d11449842b17

omni-relayer/src/workers/near.rs
frolvanya: Fixed in f218c04b15a5f0c5f72a4440b203d11449842b17

omni-relayer/src/workers/near.rs
frolvanya: Fixed in f218c04b15a5f0c5f72a4440b203d11449842b17

omni-relayer/src/utils/near.rs
frolvanya: Yep, it makes sense 
Fixed in eb3e8a5d772bfb61d88b550f239fcefe8dd560a5

omni-relayer/src/utils/near.rs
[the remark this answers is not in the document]
frolvanya: > If nonce is not none, we should handle the FinTransferEvent as well. We should initiate SignTransfer.

Something like this? e4da3b1eb0b6464aa70bfee9c16538b5491e5a7b

omni-relayer/src/config.rs
```diff
@@ -26,6 +27,7 @@ pub struct Near {
     pub rpc_url: String,
     pub token_locker_id: AccountId,
     pub credentials_path: Option<String>,
+    pub relayer_address_on_evm: OmniAddress,
```
olga24912, line 30: Why relayer_address_on_evm in Near Config not in EVM Config?

olga24912 (changes requested):

omni-relayer/src/config.rs
frolvanya: Makes sense, I'll fix it. But can we please rename it to `relayer_address`. I don't see a point of using config.**evm**.XXX_on_**evm**

frolvanya: Fixed in 068b62c17626684187b8efbe99e76265bec3e6b3

omni-relayer/src/utils/near.rs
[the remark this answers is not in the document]
olga24912: Sure, but I also want to make changes in the protocol. 

And now it is not just INIT_TRANSFER_EVENT, so, we should do some renaming. 

I think, the changes in the protocol we can do in the separate PR.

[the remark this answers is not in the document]
frolvanya: > And now it is not just INIT_TRANSFER_EVENT, so, we should do some renaming.

Fixed in 5133a5baabbbebf474b9cc567110be979252af4b

olga24912 (approved):

PR #71, opened 2024-10-11
feat: wait for eth light client

### LINKED ISSUE #63: [OmniBridge]: Relayer, wait for the block from Ethereum to be processed on NEAR.
After an event occurs on Ethereum, wait for 20 minutes until the required block appears on NEAR before submitting the proof and proceeding with further logic.

frolvanya:
closes #63


omni-relayer/src/startup/evm.rs
```diff
@@ -94,238 +58,64 @@ pub async fn start_indexer(
...
-            return;
-        }
-
+    } else if log.log_decode::<utils::evm::FinTransfer>().is_ok() {
         utils::redis::add_event(
             redis_connection,
             utils::redis::FINALIZED_TRANSFERS,
             tx_hash.to_string(),
-            serialized_claim_fee_args,
+            log,
```
olga24912, line 110: Why we don't save the block number here as well?

omni-relayer/src/workers/evm.rs
```diff
@@ -33,32 +41,117 @@ pub async fn finalize_withdraw(
...
+            {
+                let Ok(light_client_latest_block_number) =
+                    utils::near::get_eth_light_client_last_block_number(&config, &jsonrpc_client)
+                        .await
+                else {
+                    warn!("Failed to get eth light client last block number");
+                    continue;
+                };
+
+                if block_number > light_client_latest_block_number {
```
olga24912, line 55: I assume that the events are in chronological order. Then it makes sense to add a sleep and jump to the loop.

omni-relayer/src/workers/evm.rs
```diff
@@ -33,32 +41,117 @@ pub async fn finalize_withdraw(
...
+                        };
+                        let Ok(recipient) = init_log.inner.recipient.parse::<AccountId>() else {
+                            warn!(
+                                "Failed to parse recipient as AccountId: {:?}",
+                                init_log.inner.recipient
+                            );
+                            return;
+                        };
+
+                        let vaa = utils::evm::get_vaa(tx_logs, &log, &config).await;
```
olga24912, line 100: In case of wormhole we don't need to wait for block number on EthLightClient.

omni-relayer/src/startup/evm.rs
olga24912: In case of FinTransferEvent, before we are doing the "cliam_fee", we should check the bock number as well.

olga24912 (changes requested):

omni-relayer/src/config.rs
```diff
@@ -46,6 +46,7 @@ pub struct Near {
     pub rpc_url: String,
     pub token_locker_id: AccountId,
     pub credentials_path: Option<String>,
+    pub eth_light_client: AccountId,
```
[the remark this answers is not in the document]
frolvanya: Fixed in ad692b6

omni-relayer/src/startup/evm.rs
frolvanya: Fixed in 998b21e

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in 68821cd

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in 8b469bc

omni-relayer/src/workers/near.rs
```diff
@@ -178,7 +183,24 @@ pub async fn claim_fee(redis_client: redis::Client, connector: Arc<OmniConnector
...
+                    continue;
+                };
+
+                if block_number > light_client_latest_block_number {
+                    tokio::time::sleep(tokio::time::Duration::from_secs(
+                        utils::redis::SLEEP_TIME_AFTER_EVENTS_PROCESS_SECS,
+                    ))
+                    .await;
+                    continue;
+                }
```
olga24912, line 202: In this function, we sometimes can use a wormhole, but we don't.

olga24912 (changes requested):

omni-relayer/src/workers/near.rs
frolvanya: Fixed in 1a8def1c8769bde5bacf888685ea0f803e717cef

olga24912 (approved):

PR #74, opened 2024-10-14
feat: EVM locker implementation

karim-en:
- Implemented EVM locker
- Changed the interface of the `EVM::initTransfer`, so it now accepts the token address instead of the token ID and message instead of `near_account_id:msg`.
- Map the token address and ID in both directions on the Near side to get the token ID by the address.
- Added the `SolAddress` type with validation.

Todo in follow-up PRs:
- Rename the factory contract and move the hardhat project to evm dir.
- Implement factory on Near

### DIFF
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -18,6 +18,9 @@ contract BridgeTokenFactory is
     AccessControlUpgradeable,
     SelectivePausableUpgradable
 {
+    // We removed ProofConsumer from the list of parent contracts and added this gap
+    // to preserve storage layout when upgrading to the new contract version.
+    uint256[54] private __gap;
     using SafeERC20 for IERC20;
     mapping(address => string) public ethToNearToken;
     mapping(string => address) public nearToEthToken;
@@ -27,6 +30,7 @@ contract BridgeTokenFactory is
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
+    uint256[3] private __gapForRemovedFields;
     mapping(uint128 => bool) public completedTransfers;
     mapping(uint128 => bool) public claimedFee;
     uint128 public currentNonce; 
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -27,7 +27,6 @@ contract BridgeTokenFactory is
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
-    uint256[3] private __gapForRemovedFields;
     mapping(uint128 => bool) public completedTransfers;
     mapping(uint128 => bool) public claimedFee;
     uint128 public currentNonce; 
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -33,10 +33,8 @@ contract BridgeTokenFactory is
 
     bytes32 public constant PAUSABLE_ADMIN_ROLE = keccak256("PAUSABLE_ADMIN_ROLE");
     uint constant UNPAUSED_ALL = 0;
-    uint constant PAUSED_BURN_TOKEN = 1 << 0;
-    uint constant PAUSED_MINT_TOKEN = 1 << 1;
-    uint constant PAUSED_LOCK_TOKEN = 1 << 2;
-    uint constant PAUSED_UNLOCK_TOKEN = 1 << 3;
+    uint constant PAUSED_INIT_TRANSFER = 1 << 0;
+    uint constant PAUSED_FIN_TRANSFER = 1 << 1;
 
     error InvalidSignature();
     error NonceAlreadyUsed(uint256 nonce);
@@ -125,10 +123,10 @@ contract BridgeTokenFactory is
         );
     }
 
-    function mintToken(
+    function finTransfer(
         bytes calldata signatureData, 
         BridgeTypes.FinTransferPayload calldata payload
-    ) payable external whenNotPaused(PAUSED_MINT_TOKEN) {
+    ) payable external whenNotPaused(PAUSED_FIN_TRANSFER) {
         if (completedTransfers[payload.nonce]) {
             revert NonceAlreadyUsed(payload.nonce);
         }
@@ -179,7 +177,7 @@ contract BridgeTokenFactory is
         uint128 nativeFee,
         string calldata recipient,
         string calldata message
-    ) payable external whenNotPaused(PAUSED_BURN_TOKEN) {
+    ) payable external whenNotPaused(PAUSED_INIT_TRANSFER) {
         currentNonce += 1;
         if (fee >= amount) {
             revert InvalidFee();
@@ -247,7 +245,7 @@ contract BridgeTokenFactory is
     }
 
     function pauseAll() external onlyRole(PAUSABLE_ADMIN_ROLE) {
-        uint flags = PAUSED_MINT_TOKEN | PAUSED_BURN_TOKEN | PAUSED_LOCK_TOKEN | PAUSED_UNLOCK_TOKEN;
+        uint flags = PAUSED_FIN_TRANSFER | PAUSED_INIT_TRANSFER;
         _pause(flags);
     }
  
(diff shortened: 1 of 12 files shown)


karim-en:
I think it is better to open separate PR for these changes:

- Removed the unused whitelist logic to improve code readability and contract size.
- Move types to separate libraries to improve code readability.

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -5,104 +5,43 @@ import {AccessControlUpgradeable} from "@openzeppelin/contracts-upgradeable/acce
...
-    bool private _isWhitelistModeEnabled;
+    mapping(address => string) public ethToNearToken;
+    mapping(string => address) public nearToEthToken;
+    mapping(address => bool) public isBridgeToken;
 
     address public tokenImplementationAddress;
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
+    uint256[3] private __gapForRemovedFields;
```
kiseln, line 34: This implies we intend to upgrade some of the contracts? I thought we will deploy a new one

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -241,33 +169,118 @@ contract BridgeTokenFactory is
...
+        uint256 value
+    ) internal virtual {}
+
+    function lockToken(
+        address tokenAddress,
+        uint128 amount,
+        uint128 fee,
+        uint128 nativeFee,
+        string calldata recipient,
+        string calldata message
```
kiseln, line 214: What is the message and why do we have it here but not for `burn`

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -299,113 +312,24 @@ contract BridgeTokenFactory is
...
-        _pause(pausedFlags() | PAUSED_FIN_TRANSFER);
-    }
-
-    function pauseInitTransfer() external onlyRole(PAUSABLE_ADMIN_ROLE) {
-        _pause(pausedFlags() | PAUSED_INIT_TRANSFER);
-    }
-
     function pauseAll() external onlyRole(PAUSABLE_ADMIN_ROLE) {
-        uint flags = PAUSED_FIN_TRANSFER | PAUSED_INIT_TRANSFER;
+        uint flags = PAUSED_MINT_TOKEN | PAUSED_BURN_TOKEN;
```
kiseln, line 320: Should we keep all four flags here?

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -299,113 +312,24 @@ contract BridgeTokenFactory is
...
         address implementation
     ) external onlyRole(DEFAULT_ADMIN_ROLE) {
-        require(_isBridgeToken[_nearToEthToken[nearTokenId]], "ERR_NOT_BRIDGE_TOKEN");
-        BridgeToken proxy = BridgeToken(payable(_nearToEthToken[nearTokenId]));
+        require(isBridgeToken[nearToEthToken[nearTokenId]], "ERR_NOT_BRIDGE_TOKEN");
+        BridgeToken proxy = BridgeToken(payable(nearToEthToken[nearTokenId]));
         proxy.upgradeToAndCall(implementation, bytes(""));
     }
 
-    function enableWhitelistMode() external onlyRole(DEFAULT_ADMIN_ROLE) {
```
kiseln, line 356: Happy about removing these

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: This one should also be added for `burnToken`, to replace the `account_id:msg` format

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: Yeah, I think it is better to deploy a new one, but we should be careful when changing the owner of the already deployed tokens. Also, we need to make sure that the old factory has this functionality.

kiseln: Why do we need a gap then? I think we'll deploy a new contract and then phase out the old ones?

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
kiseln: Then I guess is a question whether we want to separate lock/burn and unlock/mint functionality or have all under init and finalize. The interfaces look to be the same so I think I prefer keeping initializeTransfer and finalizeTransfer.

karim-en: Yeh, we can try to unify it, but we need to consider that the bind token isn't obligatory for now, so we should be careful what kind of token address we pass.

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -142,10 +148,14 @@ contract BridgeTokenFactory is
...
-        require(isBridgeToken[nearToEthToken[payload.token]], "ERR_NOT_BRIDGE_TOKEN");
-        BridgeToken(nearToEthToken[payload.token]).mint(payload.recipient, payload.amount);
-
         completedTransfers[payload.nonce] = true;
+        address tokenAddress = nearToEthToken[payload.token];
+        
+        if (tokenAddress != address(0)) {
+            BridgeToken(tokenAddress).mint(payload.recipient, payload.amount);
+        } else {
+            IERC20(tokenAddress).safeTransfer(payload.recipient, payload.amount);
```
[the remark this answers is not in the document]
karim-en: f7faf2a1001d79fd51ff4bebb1c4ed87cc949fc5

evm/bridge-token-factory/contracts/BridgeTypes.sol
```diff
@@ -27,16 +27,16 @@ library BridgeTypes {
...
         uint128 nativeFee,
-        string recipient
+        string recipient,
+        string message
     );
 
     event FinTransfer(
         uint128 indexed nonce,
-        string token,
+        address tokenAddress,
```
olga24912, line 39: The relayer should be fixed as well: https://github.com/Near-One/omni-bridge/blob/main/omni-relayer/src/startup/eth.rs#L41C1-L41C22

near/omni-types/src/evm/events.rs
```diff
@@ -14,11 +14,11 @@ sol! {
         address indexed sender,
         address indexed tokenAddress,
         uint128 indexed nonce,
-        string token,
         uint128 amount,
         uint128 fee,
         uint128 nativeTokenFee,
-        string recipient
+        string recipient,
+        string message
```
olga24912, line 21: FinTransfer in this file should be updated as well:
string token -> address tokenAddress

olga24912 (changes requested):

near/omni-types/src/evm/events.rs
karim-en: f1a089f9b1aeb39217d2882c1d440330a0d67975

evm/bridge-token-factory/contracts/BridgeTypes.sol
karim-en: The relayer is out of the scope.
CC @frolvanya

olga24912 (approved):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -161,28 +171,43 @@ contract BridgeTokenFactory is
...
+        string calldata message
     ) payable external whenNotPaused(PAUSED_INIT_TRANSFER) {
         initTransferNonce += 1;
-        require(isBridgeToken[nearToEthToken[token]], "ERR_NOT_BRIDGE_TOKEN");
         if (fee >= amount) {
             revert InvalidFee();
         }
 
-        address tokenAddress = nearToEthToken[token];
+        uint256 extensionValue = msg.value - nativeFee;
```
kiseln, line 186: Still feels like calculating value inside the extension method is more appropriate

kiseln (approved):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: Nope, the difference should be calculated here for two reasons: 
1- we could implement a refund later.
2- it works as a verification of `msg.value >= nativeFee `

PR #76, opened 2024-10-16
feat: derive eth address from private key

frolvanya:


omni-relayer/src/config.rs
```diff
@@ -13,14 +13,10 @@ fn derive_evm_address_from_private_key() -> OmniAddress {
...
     let secp = secp256k1::Secp256k1::new();
     let public_key = secp256k1::PublicKey::from_secret_key(&secp, &secret_key);
-    let public_key_bytes = public_key.serialize_uncompressed();
-    let public_key_bytes = &public_key_bytes[1..];
-
-    let result = keccak256(public_key_bytes);
-    let address = &result[12..];
+    let result = keccak256(&public_key.serialize_uncompressed()[1..]);
 
     OmniAddress::Eth(
```
karim-en, line 18: How are we going to support multiple EVMs with this setup?

karim-en (changes requested):

omni-relayer/src/config.rs
frolvanya: Good question... I guess we have 2 options:
1) By having different fields in config for each chain, but it will introduce mess later in the code during `claim_native_fee` call
2) Try to parse as eth, then as arb and then base. Which is not very efficient

Do you have any suggestions?

frolvanya: I believe that @kiseln proposed an idea of having unified enum `Evm` in `OmniAddress` that would behave same as current implementation but it would be much easier to use it

frolvanya: @karim-en What will be desired logic when Solana and other chains will be supported? We would need to check sender, determine origin chain and based on that provide a `native_fee_recipient`? If this is the case then there's no point in "simplifying" things by having one field in config, we can introduce:

```rust
pub relayer_address_on_eth: OmniAddress,
pub relayer_address_on_arb: OmniAddress,
pub relayer_address_on_base: OmniAddress,
pub relayer_address_on_sol: OmniAddress,
...
```

karim-en: The best solution will be just to use `map<ChainKind, OmniAddress>`, so you can get the origin ChainKind from the transfer message and then get the address from this map.
For the private key, you can also use the Omni address format `eth:private_key`...

karim-en (changes requested):

frolvanya:
TODO: Right now I went with simple strategy to add `on_{chain}` to the relayer. In the next PR I'll do it like @karim-en proposed by adding extended config to have mapping (chainkind -> address). This way we will have more generic implementation

karim-en (approved):

PR #78, opened 2024-10-18
EVM-Factory: Cleanup and remove whitelist

karim-en:
- Removed the unused whitelist logic to improve code readability and contract size.
- Move types to separate libraries to improve code readability.
- Fixed tests

### DIFF
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -9,30 +9,16 @@ import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "./BridgeToken.sol";
 import "./SelectivePausableUpgradable.sol";
 import "./Borsh.sol";
+import "./BridgeTypes.sol";
 
 contract BridgeTokenFactory is
     UUPSUpgradeable,
     AccessControlUpgradeable,
     SelectivePausableUpgradable
 {
-    enum WhitelistMode {
-        NotInitialized,
-        Blocked,
-        CheckToken,
-        CheckAccountAndToken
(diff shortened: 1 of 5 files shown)


evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -151,7 +66,7 @@ contract BridgeTokenFactory is
             revert InvalidSignature();
         }
 
-        require(!_isBridgeToken[_nearToEthToken[metadata.token]], "ERR_TOKEN_EXIST");
+        require(!isBridgeToken[nearToEthToken[metadata.token]], "ERR_TOKEN_EXIST");
```
kiseln, line 69: Let's also replace strings with custom errors. Btw, in 0.8.26 Solidity allows to use custom errors in require statements 
https://soliditylang.org/blog/2024/05/21/solidity-0.8.26-release-announcement/

karim-en: I'm getting this error 
```
UnimplementedFeatureError: Require with a custom error is only available using the via-ir pipeline.
Error HH600: Compilation failed
```

kiseln: It can be turned on this way in the config file https://ethereum.stackexchange.com/a/142206

In general, do we want to switch to IR pipeline? I think that's fine, it's the newer approach https://soliditylang.org/blog/2024/07/12/a-closer-look-at-via-ir/

karim-en: Ok, let's add and test it in a follow-up PR

kiseln (approved):

PR #80, opened 2024-10-18
 Add CI with build and linting 

kisialiou:
- Added simple CI with Clippy.
- Made minor changes to address points underlined by clippy.

### DIFF
--- a/near/omni-prover/evm-prover/src/lib.rs
+++ b/near/omni-prover/evm-prover/src/lib.rs
@@ -81,7 +81,7 @@ impl EvmProver {
                     .with_static_gas(VERIFY_PROOF_CALLBACK_GAS)
                     .verify_proof_callback(
                         args.proof_kind,
-                        evm_proof.log_entry_data.clone(),
+                        evm_proof.log_entry_data,
                         header.hash.ok_or("ERR_HASH_NOT_SET")?.0,
                     ),
             ))
@@ -176,7 +176,7 @@ impl EvmProver {
                     vec![]
                 } else {
                     Self::_verify_trie_proof(
-                        new_expected_root.clone(),
+                        new_expected_root,
                         key,
                         proof,
                         key_index + 1,
@@ -217,7 +217,7 @@ impl EvmProver {
                 require!(path.as_slice() == &key[key_index..key_index + path.len()]);
                 let new_expected_root = get_vec(&node, 1);
                 Self::_verify_trie_proof(
-                    new_expected_root.clone(),
+                    new_expected_root,
                     key,
                     proof,
                     key_index + path.len(),
--- a/near/omni-prover/evm-prover/src/lib.rs
+++ b/near/omni-prover/evm-prover/src/lib.rs
@@ -46,9 +46,10 @@ impl EvmProver {
     ///
     /// This function will panic in the following situations:
     /// - If the log entry at the specified index doesn't match the decoded log entry.
+    #[allow(clippy::needless_pass_by_value)]
     #[handle_result]
-    pub fn verify_proof(&self, #[serializer(borsh)] input: &[u8]) -> Result<Promise, String> {
-        let args = EvmVerifyProofArgs::try_from_slice(input).map_err(|_| "ERR_PARSE_ARGS")?;
+    pub fn verify_proof(&self, #[serializer(borsh)] input: Vec<u8>) -> Result<Promise, String> {
+        let args = EvmVerifyProofArgs::try_from_slice(&input).map_err(|_| "ERR_PARSE_ARGS")?;
 
         let evm_proof = args.proof;
         let header: BlockHeader = rlp::decode(&evm_proof.header_data).map_err(|e| e.to_string())?;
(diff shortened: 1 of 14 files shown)


karim-en (commented):
The contract isn't compilable anymore, please run `RUSTFLAGS='-C link-arg=-s' cargo build --target wasm32-unknown-unknown --release`

kisialiou:
> The contract isn't compilable anymore, please run `RUSTFLAGS='-C link-arg=-s' cargo build --target wasm32-unknown-unknown --release`

I fixed this and added an additional job which checks that the project is built with these flags and under the target.

olga24912 (approved):

near/omni-prover/wormhole-omni-prover-proxy/src/lib.rs
```diff
@@ -31,11 +31,12 @@ impl WormholeOmniProverProxy {
         Self { prover_account }
     }
 
+    #[allow(clippy::needless_pass_by_value)]
     pub fn verify_proof(&self, #[serializer(borsh)] input: Vec<u8>) -> Promise {
         let args = WormholeVerifyProofArgs::try_from_slice(&input)
             .unwrap_or_else(|_| env::panic_str("ErrorOnArgsParsing"));
 
-        env::log_str(&format!("{}", args.vaa));
+        env::log_str(&args.vaa.to_string());
```
karim-en, line 39: It is already string, so I would prefer to use `.clone()`

near/omni-types/src/evm/events.rs
```diff
@@ -40,12 +40,12 @@ sol! {
...
 pub fn parse_evm_event<T: SolEvent, V: TryFromLog<Log<T>>>(
     chain_kind: ChainKind,
-    log_rlp: Vec<u8>,
+    log_rlp: &[u8],
 ) -> Result<V, String>
 where
     <V as TryFromLog<Log<T>>>::Error: std::fmt::Display,
 {
-    let rlp_decoded = Log::decode(&mut log_rlp.as_slice()).map_err(stringify)?;
+    let rlp_decoded = Log::decode(&mut log_rlp.to_vec().as_slice()).map_err(stringify)?;
```
karim-en, line 48: And extra clone has been introduced here due to `to_vec()`, the `log_rlp` should be moved instead

karim-en (changes requested):

near/omni-types/src/evm/events.rs
kisialiou: Are you sure? Initially, `log_rlp` was passed by value. Now, it's passed by immutable reference. Is moving it with mutable reference safe if it was not expected earlier?

karim-en: Yes, it looks like Clippy didn't understand the flow correctly. the `log_rlp` should be moved as a `vec` because we need to modify it in the `Log::decode`

kisialiou: I would simply return `Vec` here then

near/omni-prover/wormhole-omni-prover-proxy/src/lib.rs
karim-en, line 39:
```suggestion
        env::log_str(&args.vaa);
```

karim-en (changes requested):

near/omni-prover/evm-prover/src/lib.rs
```diff
@@ -164,17 +171,17 @@ impl EvmProver {
...
             } else {
                 let new_expected_root = get_vec(&node, key[key_index] as usize);
-                if !new_expected_root.is_empty() {
+                if new_expected_root.is_empty() {
+                    // not included in proof
+                    vec![]
+                } else {
                     Self::_verify_trie_proof(
-                        new_expected_root,
+                        new_expected_root.clone(),
```
karim-en, line 179: extra clone

near/omni-prover/evm-prover/src/lib.rs
```diff
@@ -210,7 +217,7 @@ impl EvmProver {
                 require!(path.as_slice() == &key[key_index..key_index + path.len()]);
                 let new_expected_root = get_vec(&node, 1);
                 Self::_verify_trie_proof(
-                    new_expected_root,
+                    new_expected_root.clone(),
```
karim-en, line 220: extra clone

karim-en (changes requested):

near/omni-prover/evm-prover/src/lib.rs
```diff
@@ -76,12 +81,13 @@ impl EvmProver {
                     .with_static_gas(VERIFY_PROOF_CALLBACK_GAS)
                     .verify_proof_callback(
                         args.proof_kind,
-                        evm_proof.log_entry_data,
+                        evm_proof.log_entry_data.clone(),
```
karim-en, line 84: extra clone

karim-en (changes requested):

karim-en (approved):

PR #81, opened 2024-10-21
feat: scratch implementation of fee estimation

### LINKED ISSUE #48: [OmniBridge]: Support NativeFee in the Relayer
- [x] Perform a careful calculation to determine if the fee is sufficient.
- [x] Support Native Fee in the calculation
- [x] Support retrying the check for fee sufficiency.

frolvanya:
closes #48

### DIFF
--- a/omni-relayer/src/utils/fee.rs
+++ b/omni-relayer/src/utils/fee.rs
@@ -18,7 +18,7 @@ use super::storage;
 
 #[derive(Debug, serde::Deserialize)]
 struct Metadata {
-    pub decimals: u32,
+    decimals: u32,
 }
 
 async fn get_token_decimals(jsonrpc_client: &JsonRpcClient, token: &AccountId) -> Result<u32> {
@@ -105,17 +105,17 @@ pub async fn is_fee_sufficient(
             .await?
         }
         OmniAddress::Near(address) => {
-            if !storage::has_storage_deposit(jsonrpc_client, token, &address.parse::<AccountId>()?)
+            if storage::has_storage_deposit(jsonrpc_client, token, &address.parse::<AccountId>()?)
                 .await?
             {
+                0.0
+            } else {
                 calculate_price(
                     storage::NEP141_STORAGE_DEPOSIT,
                     recipient_symbol,
                     recipient_token_decimals,
                 )
                 .await?
-            } else {
-                0.0
             }
         }
         OmniAddress::Sol(_) => todo!(),
--- a/omni-relayer/src/utils/near.rs
+++ b/omni-relayer/src/utils/near.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use log::{info, warn};
+use log::info;
 
 use near_jsonrpc_client::{
     methods::{self, block::RpcBlockRequest},
@@ -64,7 +64,6 @@ pub async fn get_eth_light_client_last_block_number(
 pub async fn handle_streamer_message(
     config: &config::Config,
     redis_connection: &mut redis::aio::MultiplexedConnection,
-    jsonrpc_client: &JsonRpcClient,
     streamer_message: &StreamerMessage,
 ) {
     let nep_locker_event_outcomes = find_nep_locker_event_outcomes(config, streamer_message);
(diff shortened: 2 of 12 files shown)


omni-relayer/src/utils/fee.rs
```diff
@@ -32,49 +40,122 @@ pub async fn get_token_decimals(jsonrpc_client: &JsonRpcClient, token: &AccountI
...
+            );
+            calculate_price(
+                config.evm.init_transfer_gas_estimation * http_provider.get_gas_price().await?,
+                sender_symbol,
+                sender_token_decimals,
+            )
+            .await?
+        }
+        OmniAddress::Near(address) => {
+            if !storage::has_storage_deposit(jsonrpc_client, token, &address.parse::<AccountId>()?)
```
attila0x2A, line 107: @frolvanya If sender is sending the token is it possible for them not to have storage deposit? (Wouldn't it be required to hold the token in the first place?)

omni-relayer/src/utils/fee.rs
```diff
@@ -32,49 +40,124 @@ pub async fn get_token_decimals(jsonrpc_client: &JsonRpcClient, token: &AccountI
...
+    let (recipient_symbol, recipient_token_decimals) =
+        get_symbol_and_decimals_by_chain(recipient.get_chain());
+
+    let native_fee_usd =
+        calculate_price(fee.native_fee.0, sender_symbol, sender_token_decimals).await?;
+
+    let fee_token_decimals = get_token_decimals(jsonrpc_client, token).await?;
+    let token_fee_usd = calculate_price(fee.fee.0, sender_symbol, fee_token_decimals).await?;
+
+    let expected_sender_fee_usd = match sender {
```
olga24912, line 90: This match isn't needed at all. The user calls init_function themselves and pays for the gas.

omni-relayer/src/utils/near.rs
```diff
@@ -112,6 +110,18 @@ pub async fn handle_streamer_message(
                     }
                 }
             }
+            Nep141LockerEvent::UpdateFeeEvent {
+                ref transfer_message,
+            } => {
+                // TODO: check if transcation is already in redis (in any other case we can ignore it, because there's nothing to update)
+                utils::redis::add_event(
```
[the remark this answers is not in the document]
frolvanya: Yep, I already relocated fee check closer to the actual function call. I'll open PR today/tomorrow

omni-relayer/src/utils/fee.rs
```diff
@@ -32,49 +40,124 @@ pub async fn get_token_decimals(jsonrpc_client: &JsonRpcClient, token: &AccountI
...
+    let (sender_symbol, sender_token_decimals) =
+        get_symbol_and_decimals_by_chain(sender.get_chain());
+    let (recipient_symbol, recipient_token_decimals) =
+        get_symbol_and_decimals_by_chain(recipient.get_chain());
+
+    let native_fee_usd =
+        calculate_price(fee.native_fee.0, sender_symbol, sender_token_decimals).await?;
+
+    let fee_token_decimals = get_token_decimals(jsonrpc_client, token).await?;
+    let token_fee_usd = calculate_price(fee.fee.0, sender_symbol, fee_token_decimals).await?;
```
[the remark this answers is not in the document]
frolvanya: I guess it's just a bad naming. sender_symbol is a symbol for origin token. For example, during NEAR->ETH transfer sender_symbol will be `"near"`

omni-relayer/src/utils/fee.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/81/commits/0dec505603bc60e1b46585376af33f46d61b2d6c

omni-relayer/src/utils/fee.rs
frolvanya: Fixed it in https://github.com/Near-One/omni-bridge/pull/81/commits/0dec505603bc60e1b46585376af33f46d61b2d6c. In fact we can completely ignore all charges that come from transfer initialisation since user pays for it

omni-relayer/src/utils/near.rs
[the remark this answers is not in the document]
frolvanya: Will be fixed in https://github.com/Near-One/omni-bridge/pull/105

frolvanya:
In the future relayer will depend on Bridge Indexer API to estimate fee, so we can leave the current implementation for now and replace it with API call later

https://github.com/Near-One/bridge-indexer-rs/pull/70

olga24912 (approved):

PR #89, opened 2024-10-22
fix: add prefix to payload messages

karim-en:


evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -55,6 +55,7 @@ contract BridgeTokenFactory is
 
     function deployToken(bytes calldata signatureData, BridgeTypes.MetadataPayload calldata metadata) payable external returns (address) {
         bytes memory borshEncoded = bytes.concat(
+            bytes("\x01"),
```
kiseln, line 58: Should we have an enum instead?

near/omni-types/src/lib.rs
```diff
@@ -323,8 +323,16 @@ impl TransferMessage {
     }
 }
 
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone)]
+pub enum PayloadPrefix {
```
kiseln, line 327: I'd called it PayloadType perhaps but up to you

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: 6d94fd3f342218bab869f1623a8edc5ba03506dc

near/omni-types/src/lib.rs
karim-en: 6d94fd3f342218bab869f1623a8edc5ba03506dc

kiseln (approved):

PR #90, opened 2024-10-23
Fix existing tests and add them to CI workflow.

kisialiou:


near/omni-tests/src/lib.rs
```diff
@@ -193,7 +200,17 @@ mod tests {
             .await?
             .into_result()?;
 
-        // Add factory address
+        locker_contract
+            .call("storage_deposit")
+            .args_json(json!({
+                "account_id": relayer_account.id(),
+            }))
+            .deposit(NEP141_DEPOSIT.saturating_mul(2))
```
karim-en, line 208: Let's get the needed amount from the view methods.
`required_balance_for_account`, `required_balance_for_init_transfer` and `required_balance_for_fin_transfer`

karim-en (approved):

PR #100, opened 2024-10-31
Add unit test for nep141-locker.

kisialiou:


near/nep141-locker/src/lib.rs
```diff
@@ -968,3 +968,673 @@ impl Contract {
...
+            )],
+        );
+
+        contract.fin_transfer_callback(
+            &storage_args,
+            predecessor,
+            None, // Missing fee recipient when native fee is non-zero
+        );
+    }
+}
```
karim-en, line 1640: Let's move the tests to separate file

karim-en (changes requested):

karim-en (approved):

PR #106, opened 2024-11-06
fix: fix Wormhole prover

karim-en:
- Fixed decode abi issue
- Fixed the return type from json to proof
- Fixed decoding `emitter_address` 
- Fixed gas error

### DIFF
--- a/near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs
+++ b/near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs
@@ -113,6 +113,7 @@ impl ParsedVAA {
 
 sol! {
     struct InitTransferWh {
+        uint8 messageType;
         address sender;
         address tokenAddress;
         uint128 nonce;
@@ -124,6 +125,7 @@ sol! {
     }
 
     struct FinTransferWh {
+        uint8 messageType;
         string token;
         uint128 amount;
         string recipient;
@@ -131,6 +133,7 @@ sol! {
     }
 
     struct DeployTokenWh {
+        uint8 messageType;
         string token;
         address tokenAddress;
     }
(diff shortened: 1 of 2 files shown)


frolvanya (approved):

near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs
```diff
@@ -140,9 +143,8 @@ impl TryInto<InitTransferMessage> for ParsedVAA {
     type Error = String;
 
     fn try_into(self) -> Result<InitTransferMessage, String> {
-        let data: &[u8] = &self.payload[1..];
-        let transfer = InitTransferWh::abi_decode(data, true).map_err(stringify)?;
-
+        let transfer =
```
olga24912, line 146: Should we also check here messageType?

olga24912 (approved):

near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs
frolvanya: No, since `InitTransferWh` and `FinTransferWh` have different fields it simply won't be able to decode encoded `FinTransfer` as `InitTransferWh` or vice versa

https://github.com/Near-One/omni-bridge/blob/071ed8ec1c6e969af3ef6bbcdee518b5d49b2524/near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs#L115-L133

karim-en: it is already verified here
https://github.com/Near-One/omni-bridge/blob/071ed8ec1c6e969af3ef6bbcdee518b5d49b2524/near/omni-prover/wormhole-omni-prover-proxy/src/lib.rs#L74

olga24912: Yes, I mean this check. To me, it seems logical to check this byte inside try_into.

PR #108, opened 2024-11-07
chore: add backstage catalog entities

diegofigs:


near/omni-prover/omni-prover/.catalog-info.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: omni-prover
+  title: "Omni Prover"
+  description: |-
+    Takes proof of a Receipt from NEAR blockchain and checks if it's valid and included on-chain.
```
karim-en, line 8: This isn't correct, this prover is just an aggregator or proxy for others provers, so it jut redirect the call to evm or wormhole prover

karim-en: also the catalog is missed for wormhole prover

karim-en (changes requested):

near/omni-prover/omni-prover/.catalog-info.yaml
diegofigs: added and revised desc for omni-prover

karim-en (approved):

PR #110, opened 2024-11-11
Fix claim fee callback in case of no fee and minor naming.

kisialiou:


near/nep141-locker/src/lib.rs
```diff
@@ -716,16 +716,22 @@ impl Contract {
...
         );
 
-        ext_token::ext(token)
-            .with_static_gas(LOG_METADATA_GAS)
-            .with_attached_deposit(ONE_YOCTO)
-            .ft_transfer(fin_transfer.fee_recipient, U128(fee), None)
+        if fee > 0 {
+            PromiseOrValue::Promise(
+                ext_token::ext(token)
+                    .with_static_gas(LOG_METADATA_GAS)
```
frolvanya, line 722:
```suggestion
                    .with_static_gas(FT_TRANSFER_CALL_GAS)
```

I believe that we should use `FT_TRANSFER_CALL_GAS` instead of `LOG_METADATA_GAS`

karim-en, line 722:
```suggestion
                    .with_static_gas(FT_TRANSFER_GAS)
```

near/nep141-locker/src/lib.rs
```diff
@@ -716,16 +716,22 @@ impl Contract {
...
+                    .with_attached_deposit(ONE_YOCTO)
+                    .ft_transfer(fin_transfer.fee_recipient, U128(fee), None),
+            )
+        } else {
+            PromiseOrValue::Value(fin_transfer.nonce)
```
karim-en, line 727: This could be a little bit confusing, because we are returning nonce just in one case. So let's return the unit `()` value instead

karim-en (changes requested):

near/nep141-locker/src/lib.rs
kisialiou: We do the same thing in `fin_transfer_callback`, should I fix it as well for consistency?

near/nep141-locker/src/lib.rs
kisialiou: Fixed with FT_TRANSFER_GAS

near/nep141-locker/src/lib.rs
karim-en: the `fin_transfer_callback` is a little bit different, the nonce is generated that indicate that a new transfer has been initialized.

kisialiou: Fixed

karim-en (approved):

PR #115, opened 2024-11-14
feat: support different nonces for different chains

karim-en:
-  Introduce `destination_nonce` and track it for every chain separately
- Use the `TransferId` to to identify the transfer on all chains.
- Use `uint64` instead of `uint128` for nonces.

### DIFF
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -28,7 +28,7 @@ contract BridgeTokenFactory is
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
-    mapping(uint128 => bool) public completedTransfers;
+    mapping(bytes32 => bool) public completedTransfers;
     mapping(uint128 => bool) public claimedFee;
     uint128 public initTransferNonce; 
 
@@ -38,7 +38,8 @@ contract BridgeTokenFactory is
     uint constant PAUSED_FIN_TRANSFER = 1 << 1;
 
     error InvalidSignature();
-    error NonceAlreadyUsed(uint256 nonce);
+    error NonceAlreadyUsed(uint128 nonce);
+    error TransferAlreadyFinalised(uint8 chain, uint128 nonce);
     error InvalidFee();
 
     function initialize(
@@ -149,15 +150,23 @@ contract BridgeTokenFactory is
     ) internal virtual {}
 
     function finTransfer(
-        bytes calldata signatureData, 
+        bytes calldata signatureData,
         BridgeTypes.FinTransferPayload calldata payload
     ) payable external whenNotPaused(PAUSED_FIN_TRANSFER) {
-        if (completedTransfers[payload.nonce]) {
-            revert NonceAlreadyUsed(payload.nonce);
+        bytes32 transfer_id = bytes32(
+            (uint256(uint128(payload.origin_chain)) << 128) | payload.nonce
+        );
+
+        if (completedTransfers[transfer_id]) {
+            revert TransferAlreadyFinalised(
+                payload.origin_chain,
+                payload.nonce
+            );
         }
 
         bytes memory borshEncoded = bytes.concat(
             bytes1(uint8(BridgeTypes.PayloadType.TransferMessage)),
+            bytes1(payload.origin_chain),
             Borsh.encodeUint128(payload.nonce),
             bytes1(omniBridgeChainId),
             Borsh.encodeAddress(payload.tokenAddress),
@@ -174,7 +183,7 @@ contract BridgeTokenFactory is
             revert InvalidSignature();
         }
 
-        completedTransfers[payload.nonce] = true;
+        completedTransfers[transfer_id] = true;
         
         if (isBridgeToken[payload.tokenAddress]) {
             BridgeToken(payload.tokenAddress).mint(payload.recipient, payload.amount);
@@ -186,6 +195,7 @@ contract BridgeTokenFactory is
 
         emit BridgeTypes.FinTransfer(
             payload.nonce,
+            payload.origin_chain,
             payload.tokenAddress,
             payload.amount,
             payload.recipient,
--- a/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
+++ b/evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
@@ -28,8 +28,8 @@ contract BridgeTokenFactory is
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
-    mapping(uint128 => bool) public completedTransfers;
-    mapping(uint128 => bool) public claimedFee;
+    mapping(uint64 => bool) public completedTransfers;
+    mapping(uint64 => bool) public claimedFee;
     uint64 public currentOriginNonce; 
 
     bytes32 public constant PAUSABLE_ADMIN_ROLE = keccak256("PAUSABLE_ADMIN_ROLE");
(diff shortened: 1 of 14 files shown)


kiseln:
Trying to think of a better name for payload_nonce but it's tough. `destination_nonce`, maybe? As an opposite for origin_nonce

near/omni-types/src/prover_result.rs
```diff
@@ -3,17 +3,24 @@ use near_sdk::json_types::U128;
...
+    pub fee: Fee,
+    pub sender: OmniAddress,
+    pub msg: String,
     pub emitter_address: OmniAddress,
 }
 
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone)]
 pub struct FinTransferMessage {
-    pub nonce: U128,
+    pub payload_nonce: U128,
```
kiseln, line 22: I think maybe we should use this nonce purely to prevent replays on destination chains and exclude it from everywhere else. TransferId should be sufficient for every other use case and having two identifiers just adds confusion

near/omni-prover/wormhole-omni-prover-proxy/src/parsed_vaa.rs
```diff
@@ -135,7 +135,8 @@ struct FinTransferWh {
     token_address: OmniAddress,
     amount: u128,
     recipient: String,
```
olga24912, line 137: Could we rename it to make it more consistent?

```suggestion
    fee_recipient: String,
```

evm/bridge-token-factory/contracts/BridgeTypes.sol
```diff
@@ -2,10 +2,10 @@
 pragma solidity ^0.8.24;
 
 library BridgeTypes {
-    struct FinTransferPayload {
-        uint128 nonce;
-        uint128 origin_nonce;
+    struct TransferMessagePayload {
+        uint128 destination_nonce;
```
kiseln: Should probably use camelCase for solidity

evm/bridge-token-factory/contracts/BridgeTypes.sol
```diff
@@ -2,10 +2,10 @@
 pragma solidity ^0.8.24;
 
 library BridgeTypes {
-    struct FinTransferPayload {
-        uint128 nonce;
-        uint128 origin_nonce;
+    struct TransferMessagePayload {
```
kiseln, line 5: I personally liked FinTransferPayload

evm/bridge-token-factory/contracts/BridgeTypes.sol
karim-en: Outdated

evm/bridge-token-factory/contracts/BridgeTypes.sol
karim-en: Let's have consistent naming, so it will be easier to search it in the code. @olga24912 WDYT?

evm/bridge-token-factory/contracts/BridgeTypes.sol
kiseln: my bad

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -245,10 +245,10 @@ contract BridgeTokenFactory is
     ) internal virtual {}
 
     function claimNativeFee(bytes calldata signatureData, BridgeTypes.ClaimFeePayload memory payload) external {
-        bytes memory borshEncodedNonces = Borsh.encodeUint32(uint32(payload.nonces.length));
+        bytes memory borshEncodedNonces = abi.encodePacked(Borsh.encodeUint32(uint32(payload.nonces.length)));
```
kiseln, line 248: why do we need encodePacked?

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -28,17 +28,17 @@ contract BridgeTokenFactory is
     address public nearBridgeDerivedAddress;
     uint8 public omniBridgeChainId;
 
-    mapping(uint128 => bool) public completedTransfers;
-    mapping(uint128 => bool) public claimedFee;
-    uint128 public initTransferNonce; 
+    mapping(uint64 => bool) public completedTransfers;
+    mapping(uint64 => bool) public claimedFee;
```
kiseln, line 32: Can be removed?

evm/bridge-token-factory/contracts/BridgeTypes.sol
```diff
@@ -20,15 +20,15 @@ library BridgeTypes {
     }
 
     struct ClaimFeePayload {
```
kiseln, line 22: Can be removed?

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: Yep, in next [PR](https://github.com/Near-One/omni-bridge/pull/117)

evm/bridge-token-factory/contracts/BridgeTypes.sol
karim-en: Yep, in next [PR](https://github.com/Near-One/omni-bridge/pull/117)

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: I've changes the borsh encoding api, so it returns fixed bytes now.
Anyway this code will be removed.

kiseln (approved):

evm/bridge-token-factory/contracts/BridgeTypes.sol
olga24912: I also like FinTransferPayload

karim-en: But it is actually an init transfer payload 😄

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
```diff
@@ -242,10 +245,10 @@ contract BridgeTokenFactory is
     ) internal virtual {}
 
     function claimNativeFee(bytes calldata signatureData, BridgeTypes.ClaimFeePayload memory payload) external {
-        bytes memory borshEncodedNonces = Borsh.encodeUint32(uint32(payload.nonces.length));
+        bytes memory borshEncodedNonces = abi.encodePacked(Borsh.encodeUint32(uint32(payload.nonces.length)));
 
         for (uint i = 0; i < payload.nonces.length; ++i) {
-            uint128 nonce = payload.nonces[i];
+            uint64 nonce = payload.nonces[i];
```
olga24912, line 251: And I already forgot, is it destinationNonce or  originNonce? Anyway, should we also check the chain_id?

near/omni-types/src/lib.rs
```diff
@@ -360,16 +376,35 @@ impl Fee {
...
+    PartialEq,
+    Eq,
+    Default,
+    Copy,
+)]
+pub struct TransferId {
+    // The origin chain kind
+    pub chain: ChainKind,
+    // The transfer nonce that maintained on the source chain
+    pub nonce: Nonce,
```
olga24912, line 395: I prefer to make it explicit as well:

```suggestion
    pub origin_nonce: Nonce,
```

olga24912 (approved):

evm/bridge-token-factory/contracts/BridgeTokenFactory.sol
karim-en: It doesn't matter, will be removed

near/omni-types/src/lib.rs
karim-en: a9cb73ce0b057e2e194c5865f7d262c3404bea47

PR #116, opened 2024-11-14
fix: Remove excess storage deposit in test

### LINKED ISSUE #94: [OmniBridge]: `required_balance_for_fin_transfer` returns the insufficient amount
The correct fix would allow the integration test (`near/omni_tests/src/lib.rs`) to run correctly without the lines 223-224.

r-near:
With https://github.com/Near-One/omni-bridge/pull/117, we no longer need this adjustment for extra NEAR.

Closes #94

### DIFF
--- a/near/omni-tests/src/fin_transfer.rs
+++ b/near/omni-tests/src/fin_transfer.rs
@@ -196,8 +196,6 @@ mod tests {
             .saturating_mul(storage_deposit_accounts.len() as u128)
             .saturating_add(required_balance_for_fin_transfer)
             .saturating_add(NearToken::from_yoctonear(2))
-            // This is a temporary value to cover the issue with incorrect storage deposit calculation
-            .saturating_add(NearToken::from_yoctonear(1250000000000000000000));
 
         let storage_deposit_actions = storage_deposit_accounts
             .iter()
--- a/near/omni-tests/src/fin_transfer.rs
+++ b/near/omni-tests/src/fin_transfer.rs
@@ -195,7 +195,7 @@ mod tests {
         let required_deposit_for_fin_transfer = NEP141_DEPOSIT
             .saturating_mul(storage_deposit_accounts.len() as u128)
             .saturating_add(required_balance_for_fin_transfer)
-            .saturating_add(NearToken::from_yoctonear(2))
+            .saturating_add(NearToken::from_yoctonear(2));
 
         let storage_deposit_actions = storage_deposit_accounts
             .iter()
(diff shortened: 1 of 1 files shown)


karim-en:
LGTM, but potentially this will be out of dated due to the changes in this [PR](https://github.com/Near-One/omni-bridge/pull/117)

r-near:
> LGTM, but potentially this will be out of dated due to the changes in this [PR](https://github.com/Near-One/omni-bridge/pull/117)

Yep! Looks like that PR fixed the issue. Awesome to see, I just updated the PR to remove the extraneous storage deposit.

near/omni-tests/src/lib.rs
```diff
@@ -219,9 +219,7 @@ mod tests {
         let required_deposit_for_fin_transfer = NEP141_DEPOSIT
             .saturating_mul(storage_deposit_accounts.len() as u128)
             .saturating_add(required_balance_for_fin_transfer)
-            .saturating_add(NearToken::from_yoctonear(2))
-            // This is a temporary value to cover the issue with incorrect storage deposit calculation
-            .saturating_add(NearToken::from_yoctonear(1250000000000000000000));
+            .saturating_add(NearToken::from_yoctonear(2));
```
karim-en, line 222: Let's improve the UX and add these two yocto to the `pub fn required_balance_for_fin_transfer(&self) -> NearToken`

karim-en (changes requested):

near/omni-tests/src/lib.rs
r-near: Thanks to @olga24912 this change was made in #138. I've updated the PR to remove the excess storage deposit.

near/omni-tests/src/fin_transfer.rs
```diff
@@ -195,9 +195,7 @@ mod tests {
         let required_deposit_for_fin_transfer = NEP141_DEPOSIT
             .saturating_mul(storage_deposit_accounts.len() as u128)
             .saturating_add(required_balance_for_fin_transfer)
-            .saturating_add(NearToken::from_yoctonear(2))
-            // This is a temporary value to cover the issue with incorrect storage deposit calculation
-            .saturating_add(NearToken::from_yoctonear(1250000000000000000000));
+            .saturating_add(NearToken::from_yoctonear(2));
```
olga24912, line 198: This line we can remove as well. Now it is included in required_balance_for_fin_transfer. 
https://github.com/Near-One/omni-bridge/pull/138/files

olga24912 (changes requested):

karim-en (approved):

PR #117, opened 2024-11-14
feat: claim native fee on NEAR side

karim-en:
Claim native tokens on NEAR side, by minting native tokens instead of claiming fee in the source chain.

### DIFF
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -24,9 +24,8 @@ use omni_types::near_events::Nep141LockerEvent;
 use omni_types::prover_args::VerifyProofArgs;
 use omni_types::prover_result::ProverResult;
 use omni_types::{
-    BasicMetadata, ChainKind, ClaimNativeFeePayload, Fee, InitTransferMsg, MetadataPayload,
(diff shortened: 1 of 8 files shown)


kiseln:
Looks good but I don't see where we prevent claiming the fee twice

olga24912:
> Looks good but I don't see where we prevent claiming the fee twice

In this line: https://github.com/Near-One/omni-bridge/blob/main/near/nep141-locker/src/lib.rs#L607

kiseln (approved):

near/nep141-locker/src/lib.rs
```diff
@@ -625,32 +557,21 @@ impl Contract {
...
                     .transfer(NearToken::from_yoctonear(message.fee.native_fee.0));
             } else {
-                let required_balance = self.update_fin_transfer(
-                    &message.get_transfer_id(),
-                    &Some(NativeFee {
-                        amount: message.fee.native_fee,
-                        recipient: native_fee_recipient.clone(),
-                    }),
-                );
+                let native_token_address = OmniAddress::new_zero(message.get_origin_chain())
```
olga24912, line 564: new_zero -> get_native_token_address. 

It might not be zero if we use wrapped tokens.

olga24912 (approved):

PR #121, opened 2024-11-15
Add integration tests for init on NEAR.

kisialiou:
Structure:
- `TestEnv` - setup of the contracts before interactions within the major workflow;
- `init_transfer_flow_on_near` - general parametrized interaction between contracts within the major workflow;
- Tests themselves (`#[tokio::test]`) are different versions of (cases for) the major workflow.


karim-en:
@kisialiou can you please merge main into it?

kisialiou:
> LGTM! But I've noticed one drawback of mocking, is that provers aren't covered by the tests, so we need to cover the provers separately. For the EVM prover you can mock the light client, but for the wormhole you will need to run the tests in the testnet environment instead of sandbox.

Ok, I will work on this after the tests for omni token then.

PR #136, opened 2024-11-21
fix: mint native fee on fin_transfer

karim-en:
- Refactored the storage deposit arguments.
- Added verification for of storage deposit for the native token.
- Removed the hardcoded `NEP141_DEPOSIT` amount, the relayer now needs to set this value.

### DIFF
--- a/near/nep141-locker/src/lib.rs
+++ b/near/nep141-locker/src/lib.rs
@@ -17,7 +17,7 @@ use near_sdk::{
     PanicOnDefault, Promise, PromiseError, PromiseOrValue, PromiseResult,
 };
 use omni_types::locker_args::{
-    BindTokenArgs, ClaimFeeArgs, DeployTokenArgs, FinTransferArgs, StorageDepositArgs,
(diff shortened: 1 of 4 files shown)


near/nep141-locker/src/lib.rs
```diff
@@ -963,6 +960,27 @@ impl Contract {
             required_balance = required_balance.saturating_add(ONE_YOCTO);
         }
 
+        if transfer_message.fee.native_fee.0 > 0 {
+            let native_token_id = self.get_native_token_id(transfer_message.get_origin_chain());
```
olga24912, line 964: Theoretically, I can perform a transfer from NEAR to NEAR via Omni Transfer, and in that case, the native_token would be NEAR, which cannot be minted. We don’t have any checks in place to prevent this.

olga24912 (changes requested):

near/nep141-locker/src/lib.rs
karim-en: Actually in this case you won't be able to provide a proof, because we don't have a Near prover.
anyway, has been added a check here 18ab2d0c86eefc6c3a4a75632f032b215ec23322

PR #138, opened 2024-11-21
Add required_balance_for_deploy_token

olga24912:


near/nep141-locker/src/storage.rs
```diff
@@ -161,24 +163,37 @@ impl Contract {
...
-            * 2;
+            .len() as u64;
 
         env::storage_byte_cost()
-            .saturating_mul((Self::get_basic_storage() + key_len + value_len).into())
+            .saturating_mul((2 * (Self::get_basic_storage() + key_len + value_len)).into())
+    }
+
+    pub fn required_balance_for_deploy_token(&self) -> NearToken {
+        let key_len = 64 + 4;
```
karim-en, line 189: Let's add this function, and remove `64 + 4` here and in `required_balance_for_account`
```
    fn max_key_len_of_account_id() -> u64 {
        let max_account_id: AccountId = "a".repeat(64).parse().sdk_expect("ERR_PARSE_ACCOUNT_ID");
        borsh::to_vec(&max_account_id).sdk_expect("ERR_BORSH").len() as u64
    }
```

karim-en (changes requested):

karim-en (approved):

PR #145, opened 2024-11-26
feat: add multiple EVM chains configs

karim-en:
- Added configs for Arbitrum and Base networks 
- Improved the deploy script


evm/hardhat.config.ts
```diff
@@ -31,27 +44,58 @@ task("set-metadata-ft", "Set metadata for NEP-141 tokens on the Ethereum side")
...
-    console.log(`OmniBridge deployed at ${await OmniBridge.getAddress()}`)
+    const bridgeAddress = await OmniBridge.getAddress()
+    const wormholeAddressStorageValue = await hre.ethers.provider.getStorage(bridgeAddress, 58)
+
+    console.log(`OmniBridge deployed at ${bridgeAddress}`)
     console.log(
       "Implementation address:",
       await upgrades.erc1967.getImplementationAddress(await OmniBridge.getAddress()),
     )
+    console.log(`Wormhole address ${wormholeAddressStorageValue}`)
```
olga24912, line 98: Should this match the `wormholeAddress` from the config?
Should we add an additional check to ensure it’s as expected?

evm/utils/kdf.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function uncompressedHexPointToEvmAddress(uncompressedHexPoint: string): string {
+  const addressHash = ethers.keccak256(`0x${uncompressedHexPoint.slice(2)}`)
+
+  // Ethereum address is last 20 bytes of hash (40 characters), prefixed with 0x
+  return `0x${addressHash.substring(addressHash.length - 40)}`
+}
+
+async function uncompressedHexPointToBtcAddress(
```
olga24912, line 54: Why we need this function?

evm/utils/kdf.ts
```diff
@@ -0,0 +1,95 @@
+import bs58 from "bs58"
```
olga24912, line 1: Did you write this code from scratch, or is there an original source you referenced? If there is an original source, could you add a link to it in the comments? I think it could be helpful.

kiseln (approved):

olga24912 (approved):
Thank you for the PR! The deployment script looks much cleaner now. And I learned something new—I didn’t know that network config could be extended this way.

evm/utils/kdf.ts
karim-en: This has been implemented by @kiseln in previous commits, so maybe he can share the links

kiseln: https://github.com/near-examples/near-multichain/blob/main/src/services/kdf.js

evm/hardhat.config.ts
karim-en: 1df4370a93d78233196a8178bc764d5b13687aae

evm/hardhat.config.ts
```diff
@@ -31,27 +44,58 @@ task("set-metadata-ft", "Set metadata for NEP-141 tokens on the Ethereum side")
...
     const { ethers, upgrades } = hre
+    const networkConfig = hre.network.config as HttpNetworkUserConfig
+    const omniChainId = networkConfig.omniChainId
+    const wormholeAddress = networkConfig.wormholeAddress
+
+    const mpcRootPublicKey = hre.network.name.endsWith("mainnet")
+      ? mpcRootPublicKeys.mainnet.key
+      : mpcRootPublicKeys.testnet.key
+
+    const nearBridgeDerivedAddress = await deriveEthereumAddress(
```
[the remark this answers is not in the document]
karim-en: 1df4370a93d78233196a8178bc764d5b13687aae

evm/utils/kdf.ts
karim-en: 1df4370a93d78233196a8178bc764d5b13687aae

evm/utils/kdf.ts
karim-en: 1df4370a93d78233196a8178bc764d5b13687aae

PR #146, opened 2024-11-28
feat: support of multiple chains

frolvanya:


omni-relayer/src/config.rs
```diff
@@ -21,11 +31,46 @@ fn derive_evm_address_from_private_key() -> OmniAddress {
...
+
+    let signer = LocalSigner::from_signing_key(secret_key);
+
+    OmniAddress::Base(
+        H160::try_from_slice(signer.address().as_slice())
+            .expect("Failed to create `OmniAddress` from the derived public key"),
+    )
+}
+
+fn derive_arb_address_from_private_key() -> OmniAddress {
```
olga24912, line 50: These functions look very similar. Maybe it would be better to use something like `derive_evm_address_from_private_key(chain_kind: ChainKind)`?

omni-relayer/src/config.rs
```diff
@@ -49,16 +94,47 @@ pub struct Near {
...
+    #[serde(default = "derive_base_address_from_private_key")]
+    pub relayer_address: OmniAddress,
 
-    #[serde(default = "derive_evm_address_from_private_key")]
-    pub relayer_address_on_eth: OmniAddress,
+    pub block_processing_batch_size: u64,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct Arb {
```
olga24912, line 122: It would be great to have a single EvmConfig and a mapping from ChainKind to the config. Right now, the configs are identical.

omni-relayer/example-config.toml
```diff
@@ -6,12 +6,33 @@ network = "testnet"
...
+[arb]
+rpc_http_url = "https://arbitrum-sepolia.infura.io/v3/API_KEY"
+rpc_ws_url = "wss://arbitrum-sepolia.infura.io/ws/v3/API_KEY"
+chain_id = 421_614
+bridge_token_factory_address = "0xd565f7CcE0FA1bB8DBe73FCDA281390d545f6200"
+block_processing_batch_size = 10_000
+
+[wormhole]
+api_url = "https://api.testnet.wormholescan.io/"
+
```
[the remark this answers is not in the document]
olga24912: Or maybe it would make more sense to move these chain_id values into EvmConfig.

omni-relayer/src/main.rs
```diff
@@ -96,7 +80,17 @@ async fn main() -> Result<()> {
...
+    }));
+    handles.push(tokio::spawn({
+        let config = config.clone();
+        let redis_client = redis_client.clone();
+        async move { startup::evm::start_base_indexer(config, redis_client).await }
+    }));
+    handles.push(tokio::spawn({
+        let config = config.clone();
+        let redis_client = redis_client.clone();
+        async move { startup::evm::start_arb_indexer(config, redis_client).await }
```
olga24912, line 93: It would be even better if we had a map from ChainKind to a function. This way, we could see which ChainKind values are defined and call the appropriate indexers.

omni-relayer/src/startup/evm.rs
```diff
@@ -13,33 +14,157 @@ use ethereum_types::H256;
 
 use crate::{config, utils, workers::near::FinTransfer};
 
-pub async fn start_indexer(config: config::Config, redis_client: redis::Client) -> Result<()> {
+pub async fn start_eth_indexer(config: config::Config, redis_client: redis::Client) -> Result<()> {
```
olga24912, line 17: These three functions are essentially the same. They should be merged into one, accepting chainKind as a parameter.

omni-relayer/src/utils/near.rs
```diff
@@ -103,66 +103,33 @@ pub async fn handle_streamer_message(
...
-                        redis_connection,
-                        utils::redis::NEAR_SIGN_CLAIM_NATIVE_FEE_QUEUE,
-                        transfer_message.origin_nonce.0.to_string(),
-                        log,
-                    )
-                    .await;
-                }
+                utils::redis::add_event(
+                    redis_connection,
+                    utils::redis::NEAR_SIGN_CLAIM_NATIVE_FEE_QUEUE,
```
olga24912, line 116: We shouldn't have `NEAR_SIGN_CLAIM_NATIVE_FEE_QUEUE` because we don't have sign_claim_native_fee function

olga24912 (changes requested):

omni-relayer/src/config.rs
frolvanya: I tried doing this. Unfortunately, function must not have any arguments in order to be used in a serde macro

omni-relayer/src/config.rs
frolvanya: Can you please elaborate a little bit on it? How it would look like in a toml file? We can't have a duplicate keys

omni-relayer/src/config.rs
frolvanya: Since we don't need a relayer on evm anymore (we needed relayer only for native fee claiming), then this logic is now useless and I completely removed it

omni-relayer/example-config.toml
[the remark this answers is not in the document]
frolvanya: > Or maybe it would make more sense to move these chain_id values into EvmConfig.

I was thinking about this, but decided to store it in wormhole block. Let me know if you have a strong opinion on this one

omni-relayer/src/utils/near.rs
frolvanya: Well, I completely removed `sign_claim_native_fee` logic in https://github.com/Near-One/omni-bridge/pull/146/commits/8126804275e90842252e8e047083e91fcfbaa10d. But now I'm not sure what should I do on "FinTransferEvent" on near side
https://github.com/Near-One/omni-bridge/blob/76e65901ad53c77cf6eaafc97b573a60a7f0f42e/omni-relayer/src/utils/near.rs#L111-L113

omni-relayer/src/config.rs
frolvanya: I was able to come up with something like this https://github.com/Near-One/omni-bridge/pull/146/commits/68d8e84b5dae8e48514def5d5ac7e8b1df281dcd

olga24912 (changes requested):

omni-relayer/src/startup/evm.rs
```diff
@@ -48,31 +82,33 @@ pub async fn start_indexer(config: config::Config, redis_client: redis::Client)
...
+                &filter
+                    .clone()
+                    .from_block(current_block)
+                    .to_block((current_block + block_processing_batch_size).min(latest_block)),
+            )
             .await?;
 
         for log in logs {
-            process_log(&mut redis_connection, &http_provider, log).await;
+            process_log(ChainKind::Eth, &mut redis_connection, &http_provider, log).await;
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/146/commits/4c0c2db52c8bd96c2e83f14e06bb82df3284f250

olga24912 (approved):

PR #147, opened 2024-11-28
Fast Transfers

kiseln:
## Description
Fast transfer is initiated by sending tokens to the bridge contract with a proper message for `ft_on_transfer`.

### Recipient on Near
- Tokens are transferred directly to the recipient
- During finalization tokens and fees are released back to the relayer

### Recipient on other chain
- Tokens are burned/locked and a new transfer is created for a 3rd-party chain.
- During finalization tokens are released to the relayer
- Fees are paid only after finalizing on 3rd-party chain

## Implementation
- `msg` has been changed for `ft_on_transfer` to resolve between `init_transfer` and `fast_fin_transfer`.
- Fast transfers are stored in a lookup with a key being a hash of its struct. Relayer must provide correct transfer_id, token_id, recipient, amount, fee, and msg. 
- Storage deposit is only performed if recipient is on Near in order to be able to transfer tokens to the recipient. Storage deposit is done using relayer balance instead of attached NEAR.
- Any transfers requiring native NEAR are done using wNEAR. If recipient is on Near, the tokens are unwrapped before sending.
- Only relayer that performed fast transfer can then finalize the transfer

## Concerns
- For other chain transfers sender is changed to be the relayer. I don't think this parameter is important but keeping in mind that it is different from normal transfers.
- Storage deposits need to be double checked and tested


near/omni-bridge/src/lib.rs
```diff
@@ -448,12 +453,72 @@ impl Contract {
     }
 
     #[payable]
+    pub fn fast_transfer(&mut self, #[serializer(borsh)] args: FastTransferArgs) -> Promise {
```
karim-en, line 456: The fast transfer should be initialized with ft_transfer_call

karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -425,6 +401,57 @@ impl Contract {
             )
     }
 
+    fn init_transfer(
+        &mut self,
+        sender_id: AccountId,
+        amount: U128,
+        init_transfer_msg: InitTransferMsg,
+    ) -> U128 {
+        let token_id = env::predecessor_account_id();
```
karim-en, line 410: Let's pass the `token_id` as an argument

near/omni-bridge/src/lib.rs
```diff
@@ -1005,16 +1128,39 @@ impl Contract {
...
+        // If fast transfer happened, send tokens to the relayer that executed fast transfer
+        if let Some(relayer) = recipient {
+            self.send_tokens(
+                token,
+                relayer,
+                U128(transfer_message.amount.0 - transfer_message.fee.fee.0),
+                String::new(),
+            );
+        } else {
+            required_balance = self
```
karim-en, line 1160: What if the `fast_transfer` data isn't correct, for example wrong amount or recipient, the `add_transfer_message` will fail in this case. So it should be possible to overwrite it.

karim-en (changes requested):

near/omni-bridge/src/lib.rs
kiseln: `fast_transfer` is derived from the actual transfer message in this case, unless I'm missing something?

near/omni-bridge/src/lib.rs
kiseln: Done 7aa8115d52d0c3a5f3c18c0954ddfeede6f6c83b

near/omni-bridge/src/lib.rs
```diff
@@ -162,6 +164,7 @@ pub struct Contract {
     pub factories: LookupMap<ChainKind, OmniAddress>,
     pub pending_transfers: LookupMap<TransferId, TransferMessageStorage>,
     pub finalised_transfers: LookupSet<TransferId>,
+    pub fast_transfers: LookupMap<[u8; 32], AccountId>,
```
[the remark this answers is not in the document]
kiseln: Done 7aa8115d52d0c3a5f3c18c0954ddfeede6f6c83b

near/omni-bridge/src/lib.rs
karim-en: I mean that the transfer is already inserted fast fin transfer to other chains by transfer Id, so this insert will fail if the hash of the messages are different.

kiseln: fa617c48582d9f3f3d111c9b8aca1fe53066445c

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+            msg: fast_transfer.msg,
+            destination_nonce,
+            parent_transfer_id: Some(fast_transfer.transfer_id),
+        };
+
+        required_balance = self
+            .add_transfer_message(transfer_message, relayer_id.clone())
+            .saturating_add(required_balance);
+
+        self.update_storage_balance(relayer_id, required_balance, env::attached_deposit());
```
olga24912, line 663: `env::attached_deposit()` -- this value isn’t always zero, is it?

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+        let transfer_message = TransferMessage {
+            origin_nonce: self.current_origin_nonce,
+            token: OmniAddress::Near(fast_transfer.token_id.clone()),
+            amount: fast_transfer.amount,
+            recipient: fast_transfer.recipient,
+            fee: fast_transfer.fee,
+            sender: OmniAddress::Near(relayer_id.clone()),
+            msg: fast_transfer.msg,
+            destination_nonce,
+            parent_transfer_id: Some(fast_transfer.transfer_id),
```
olga24912, line 656: Should we additionally check that the `transfer_id` has not already been finalized?

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+        sender_id: AccountId,
+        token_id: AccountId,
+        amount: U128,
+        fast_fin_transfer_msg: FastFinTransferMsg,
+    ) -> PromiseOrValue<U128> {
+        let fast_transfer = FastTransfer {
+            token_id: token_id.clone(),
+            recipient: fast_fin_transfer_msg.recipient.clone(),
+            amount: U128(amount.0 + fast_fin_transfer_msg.fee.fee.0),
+            fee: fast_fin_transfer_msg.fee,
```
olga24912, line 566: Who will ultimately receive the native fee? The Fast Transfer Relayer or the General Relayer? Are we at risk of spending the same funds twice?

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+            destination_nonce,
+            parent_transfer_id: Some(fast_transfer.transfer_id),
+        };
+
+        required_balance = self
+            .add_transfer_message(transfer_message, relayer_id.clone())
+            .saturating_add(required_balance);
+
+        self.update_storage_balance(relayer_id, required_balance, env::attached_deposit());
+    }
```
olga24912, line 664: Where do we create InitTransferEvent or something similar? How will the relayer know that this event needs to be processed? Maybe we should create a FastTransferEvent?

olga24912 (changes requested):

olga24912:
I might also consider adding an option to opt out of fast-transfer.

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+        #[serializer(borsh)] relayer_id: AccountId,
+    ) -> Promise {
+        require!(
+            Self::check_storage_balance_result(1),
+            "STORAGE_ERR: The transfer recipient is omitted"
+        );
+
+        let OmniAddress::Near(recipient) = fast_transfer.recipient.clone() else {
+            env::panic_str("ERR_INVALID_STATE")
+        };
```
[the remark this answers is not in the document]
kiseln: This would never happen because this method is only called if recipient is on Near

near/omni-bridge/src/lib.rs
kiseln: This is being discussed

near/omni-bridge/src/lib.rs
```diff
@@ -1141,6 +1320,22 @@ impl Contract {
...
+        &mut self,
+        fast_transfer: &FastTransfer,
+        relayer_id: &AccountId,
+    ) -> NearToken {
+        let storage_usage = env::storage_usage();
+        require!(
+            self.fast_transfers
+                .insert(&fast_transfer.id(), relayer_id)
+                .is_none(),
+            "Fast transfer is already performed"
```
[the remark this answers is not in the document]
kiseln: We can check it but we can't insert a fast transfer before storage deposit. So the race condition can still happen if two relayers submit fast transfer in the same block. @karim-en do we need to implement refund logic?

near/omni-bridge/src/lib.rs
kiseln: Perhaps we can allow `sign_transfer` only for the fast relayer in this case

near/omni-bridge/src/lib.rs
karim-en: The regular init transfer event should be enough since we have the parent transfer id

near/omni-types/src/lib.rs
```diff
@@ -413,6 +428,7 @@ pub struct TransferMessage {
     pub sender: OmniAddress,
     pub msg: String,
     pub destination_nonce: Nonce,
+    pub parent_transfer_id: Option<TransferId>,
```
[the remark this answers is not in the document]
karim-en: I think it is better to call it `fast_transfer_parent_id`, so it will be clear that it is related to the fast transfer.

near/omni-bridge/src/lib.rs
```diff
@@ -162,6 +164,7 @@ pub struct Contract {
     pub factories: LookupMap<ChainKind, OmniAddress>,
     pub pending_transfers: LookupMap<TransferId, TransferMessageStorage>,
     pub finalised_transfers: LookupSet<TransferId>,
+    pub fast_transfers: LookupMap<FastTransferId, AccountId>,
```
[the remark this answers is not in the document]
karim-en: I think it is better to call it fast_transfer_parent_id, so it will be clear that it is related to the fast transfer.

near/omni-types/src/lib.rs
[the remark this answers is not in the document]
karim-en: I think it is better to call it fast_transfer_parent_id, so it will be clear that it is related to the fast transfer.

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,122 @@ impl Contract {
...
+                    storage_deposit_amount: fast_fin_transfer_msg.storage_deposit_amount,
+                };
+                PromiseOrValue::Promise(
+                    Self::check_or_pay_ft_storage(
+                        &deposit_action,
+                        &mut NearToken::from_yoctonear(storage_deposit_amount),
+                    )
+                    .then(
+                        Self::ext(env::current_account_id())
+                            .with_static_gas(VERIFY_PROOF_CALLBACK_GAS)
```
karim-en, line 596: The const `VERIFY_PROOF_CALLBACK_GAS` isn't correct

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
karim-en: It doesn't make much sense to me. Even if we checked it, it is possible that this could be failed in the next block.
But I'm a little bit surprised that this implementation works, because the last promise `send_tokens` should return value, so the `resolve_transfer` will be able to process refund for the failed `ft_on_transfer`.

karim-en:
Please add integration tests to this new feature

near/omni-bridge/src/lib.rs
kiseln: If we need to log a new event then I would prefer `FastTransferEvent` and avoid including parent transfer in the struct

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
kiseln: 118a3332583d68ca30da8c9760532e2cb1fa0def

kiseln:
> @kiseln Please add integration tests to this new feature

118a3332583d68ca30da8c9760532e2cb1fa0def

near/omni-bridge/src/lib.rs
```diff
@@ -183,53 +186,26 @@ impl FungibleTokenReceiver for Contract {
...
-        required_storage_balance = required_storage_balance
-            .saturating_add(NearToken::from_yoctonear(parsed_msg.native_token_fee.0));
-
-        self.update_storage_balance(
-            sender_id,
-            required_storage_balance,
-            NearToken::from_yoctonear(0),
-        );
 
+        let token_id = env::predecessor_account_id();
```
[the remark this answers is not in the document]
kiseln: removed

near/omni-bridge/src/lib.rs
kiseln: Changed to 0

near/omni-bridge/src/lib.rs
kiseln: This is now fixed because we generated new transfer_id for fast transfer to 3rd chain

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
kiseln: I think I understood what you meant @karim-en. Is this a fix f046b2d5fc0efa9bb7df1ea1837dc05b93122919?

near/omni-bridge/src/lib.rs
karim-en: @kiseln

near/omni-bridge/src/lib.rs
karim-en: @kiseln  sounds reasonable

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
karim-en: Yeh, it looks good now.
But the naming could be confusing (in the explorer) because the nep141 implementation also has a `ft_resolve_transfer`

near/omni-bridge/src/lib.rs
```diff
@@ -875,6 +1022,14 @@ impl Contract {
     pub fn get_current_destination_nonce(&self, chain_kind: ChainKind) -> Nonce {
         self.destination_nonces.get(&chain_kind).unwrap_or_default()
     }
+
+    #[private]
+    pub fn ft_resolve_transfer(&mut self, amount: U128) -> U128 {
```
karim-en, line 1027: The naming could be confusing (in the explorer) because the nep141 implementation also has a `ft_resolve_transfer`

near/omni-bridge/src/lib.rs
```diff
@@ -875,6 +1022,14 @@ impl Contract {
...
+        match env::promise_result(0) {
+            PromiseResult::Successful(_) => U128(0),
+            PromiseResult::Failed => amount,
```
karim-en, line 1030: This refund is dangerous due to the undetermined behavior of the `ft_trnasfer_call`, so let's always return zero.

karim-en (changes requested):

near/omni-bridge/src/lib.rs
kiseln: Okay, just to confirm, if `ft_transfer_call` reverts it's sender who will lose money and not the relayer, right? So senders will be responsible for the target of `ft_transfer_call`

near/omni-types/src/lib.rs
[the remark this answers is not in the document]
kiseln: Removed it for the time being

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
kiseln: Renamed 1f38b0469619ad975744e128cc91dc404cde9b6f

near/omni-bridge/src/lib.rs
kiseln: Actually, in case recipient is on another chain we need a solution for both normal and fast transfers. So perhaps it's not quite in scope of this PR. As of now, the relayer that finalizes the transfer on the destination chain will receive the fees

near/omni-bridge/src/lib.rs
kiseln: Added 1f38b0469619ad975744e128cc91dc404cde9b6f

near/omni-bridge/src/lib.rs
kiseln: Fixed 1f38b0469619ad975744e128cc91dc404cde9b6f

near/omni-bridge/src/lib.rs
kiseln: Renamed 1f38b0469619ad975744e128cc91dc404cde9b6f

near/omni-bridge/src/lib.rs
kiseln: a3d9a72c9160d15ba1dd715cd1cde3cd3ce84817

near/omni-bridge/src/lib.rs
karim-en: Yes.
In the follow-up PR, we can try to solve the refund issue, which was mentioned multiple times in the audits.
What we should take into account is:
- The `ft_transfer_call` can fail only if the storage wasn't deposited or due to malicious contract, otherwise it returns.
- The mint call `ft_transfer_call` if there is a message, so it has the same behaviour.
- The `ft_transfer` can fail only if the storage wasn't deposited.

Since we already verify the storage deposit, then we can do the refund only if the `ft_transfer_call` returned a refund value.

karim-en: Just realized that no refund is more dangerous than I thought because it is possible to attach less gas to provoke the fail.

near/omni-bridge/src/lib.rs
```diff
@@ -55,6 +56,8 @@ const DEPLOY_TOKEN_GAS: Gas = Gas::from_tgas(50);
 const BURN_TOKEN_GAS: Gas = Gas::from_tgas(10);
 const MINT_TOKEN_GAS: Gas = Gas::from_tgas(5);
 const SET_METADATA_GAS: Gas = Gas::from_tgas(10);
+const RESOLVE_TRANSFER_GAS: Gas = Gas::from_tgas(3);
+const FAST_TRANSFER_CALLBACK_GAS: Gas = Gas::from_tgas(40);
```
karim-en, line 60: This could be not enough to do the `ft_transfer_call` in the callback

karim-en: Can you please also add tests for this flow (failed and success)

karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,147 @@ impl Contract {
...
+                    storage_deposit_amount: fast_fin_transfer_msg.storage_deposit_amount,
+                };
+                PromiseOrValue::Promise(
+                    Self::check_or_pay_ft_storage(
+                        &deposit_action,
+                        &mut NearToken::from_yoctonear(storage_deposit_amount),
+                    )
+                    .then(
+                        Self::ext(env::current_account_id())
+                            .with_static_gas(FAST_TRANSFER_CALLBACK_GAS)
```
karim-en, line 596:
```suggestion
                            .with_static_gas(FAST_TRANSFER_CALLBACK_GAS + FT_TRANSFER_CALL_GAS)
```
This could be not enough to do the `ft_transfer_call` in the callback

Can you please also add tests for this flow (failed and success)

karim-en (changes requested):

near/omni-bridge/src/lib.rs
karim-en: This can be reduced if the  `FT_TRANSFER_CALL_GAS` has been added

near/omni-bridge/src/lib.rs
karim-en: @kiseln

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
olga24912: In any case, the error could have occurred during the previous check. In that case, the tokens would already be burned, and we wouldn’t roll back the operation. The balance wouldn’t match.

near/omni-bridge/src/lib.rs
```diff
@@ -517,11 +547,147 @@ impl Contract {
...
+        let destination_nonce =
+            self.get_next_destination_nonce(fast_transfer.recipient.get_chain());
+        self.current_origin_nonce += 1;
+
+        let transfer_message = TransferMessage {
+            origin_nonce: self.current_origin_nonce,
+            token: OmniAddress::Near(fast_transfer.token_id.clone()),
+            amount: fast_transfer.amount.clone(),
+            recipient: fast_transfer.recipient.clone(),
+            fee: fast_transfer.fee.clone(),
```
olga24912, line 669: What about creating an invalid TransferMessage with a very very large fee? Finalizing it on the destination chain, and then claiming our excessively large fee. In that case, a lot of Ether, for example, would be minted for the relayer without reason.

olga24912 (changes requested):

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
kiseln: Actually if it's a bridged token (the only one we burn) bridge balance would be zero. Meaning if relayer sends an incorrect transaction the tokens will be burned and the refund to the relayer will fail. There will be no loss of funds from the bridge perspective, right? @karim-en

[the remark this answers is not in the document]
kiseln: Added tests and a fix here 88a978119074c7f4a4fc8cc18dd1f640b16eccc1
Not sure if more elegant solution is possible

near/omni-bridge/src/lib.rs
kiseln: Fixed 88a978119074c7f4a4fc8cc18dd1f640b16eccc1

near/omni-bridge/src/lib.rs
kiseln: I added restriction on claiming fee in 333be178f37933fafefc5302f264c1b39706f4ca

near/omni-bridge/src/lib.rs
```diff
@@ -473,7 +513,7 @@ impl Contract {
             args.storage_deposit_actions.len() <= 3,
             "Invalid len of accounts for storage deposit"
```
bowenwang1996, line 514: Could we add a comment here to explain why 3 is important?

near/omni-bridge/src/lib.rs
```diff
@@ -545,11 +587,150 @@ impl Contract {
...
+                fast_transfer: fast_transfer.clone(),
+                new_transfer_id: None,
+            }
+            .to_log_string(),
+        );
+
+        self.send_tokens(
+            fast_transfer.token_id,
+            recipient,
+            U128(fast_transfer.amount.0 - fast_transfer.fee.fee.0),
```
bowenwang1996, line 681: Where do we guarantee that `fast_transfer.amount > fast_transfer.fee.fee`?

near/omni-bridge/src/lib.rs
```diff
@@ -911,6 +1112,27 @@ impl Contract {
         self.destination_nonces.get(&chain_kind).unwrap_or_default()
     }
 
+    #[private]
+    pub fn resolve_transfer(&mut self, _amount: U128) -> U128 {
+        U128(0)
+    }
```
bowenwang1996, line 1118: What is the purpose of this dummy method?

near/omni-types/src/lib.rs
```diff
@@ -503,3 +520,44 @@ pub struct BasicMetadata {
...
+    pub amount: U128,
+    pub fee: Fee,
+    pub recipient: OmniAddress,
+    pub msg: String,
+}
+
+impl FastTransfer {
+    #[allow(clippy::missing_panics_doc)]
+    pub fn id(&self) -> FastTransferId {
+        FastTransferId(utils::keccak256(&borsh::to_vec(self).unwrap()))
```
bowenwang1996, line 542: is there a reason why we use keccak instead of sha256 here?

near/omni-bridge/src/lib.rs
```diff
@@ -186,53 +191,35 @@ impl FungibleTokenReceiver for Contract {
...
-            required_storage_balance,
-            NearToken::from_yoctonear(0),
-        );
-
-        if self.deployed_tokens.contains(&token_id) {
-            ext_token::ext(token_id.clone())
-                .with_static_gas(BURN_TOKEN_GAS)
-                .burn(amount);
+        if !self.deployed_tokens.contains(&token_id) {
+            return promise_or_value;
```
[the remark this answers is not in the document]
kiseln: This is a list for tokens brought from other networks. Native nep-141 tokens will not be included

near/omni-bridge/src/lib.rs
kiseln: It's to avoid refunds as decided here https://github.com/Near-One/omni-bridge/pull/147#discussion_r1883275127

near/omni-types/src/lib.rs
kiseln: Not really. Is there a reason to use the opposite?

karim-en (commented):
I don't see a call to remove fast_transfers. Is this on purpose? As I see we should be able to remove it in the same way as we remove the transfers from pending_transfers

kiseln:
> I don't see a call to remove from the `fast_transfers`. Is this on purpose? As I see we should be able to remove it in the same way as we remove the transfers from pending_transfers

I suppose we can remove it after the fee is claimed.

near/omni-bridge/src/lib.rs
```diff
@@ -196,53 +201,35 @@ impl FungibleTokenReceiver for Contract {
...
-            recipient: parsed_msg.recipient,
-            fee: Fee {
-                fee: parsed_msg.fee,
-                native_fee: parsed_msg.native_token_fee,
-            },
-            sender: OmniAddress::Near(sender_id.clone()),
-            msg: String::new(),
-            destination_nonce,
+        let parsed_msg: BridgeOnTransferMsg =
+            serde_json::from_str(&msg).sdk_expect("ERR_PARSE_MSG");
```
karim-en, line 206: Let's try to parse in the old format if the new one fails.

karim-en (changes requested):

karim-en:
@kiseln Please fix the CI

near/omni-bridge/src/lib.rs
kiseln: e67f0bcf9c67375666e359e34e808545ae7328d8

near/omni-types/src/lib.rs
karim-en: @kiseln up

kiseln: Replaced 1f58c628b6b626b098da4d92171116c574c4cbcf

near/omni-bridge/src/lib.rs
```diff
@@ -555,11 +599,149 @@ impl Contract {
...
+            msg: fast_fin_transfer_msg.msg,
+        };
+
+        if let OmniAddress::Near(recipient) = fast_fin_transfer_msg.recipient {
+            let storage_deposit_amount = fast_fin_transfer_msg
+                .storage_deposit_amount
+                .unwrap_or_default();
+            if storage_deposit_amount > 0 {
+                self.update_storage_balance(
+                    fast_fin_transfer_msg.relayer.clone(),
```
olga24912, line 628: The relayer can be any arbitrary account; we do not check whether it matches the sender_id that transfers the tokens. As a result, we can force any arbitrary account to pay us for storage.

For example, we can make a transfer of a small amount (eps) of funds, where eps << storage_deposit, find an account with sufficient storage, and force it to top up our deposit for the token.

olga24912 (changes requested):

olga24912 (changes requested):

near/omni-bridge/src/lib.rs
kiseln: Fixed with signer_id usage

karim-en (approved):

PR #148, opened 2024-11-29
Add integration tests for OmniToken.

kisialiou:


near/omni-tests/src/omni_token.rs
```diff
@@ -0,0 +1,361 @@
...
+
+            let required_storage: NearToken = locker
+                .view("required_balance_for_deploy_token")
+                .await?
+                .json()?;
+
+            token_deploy_initiator
+                .call(locker.id(), "deploy_token")
+                .args_borsh(get_test_deploy_token_args(
+                    ChainKind::Eth,
```
olga24912, line 131: Here, ChainKind should match the one used in init_token_address, and it’s not always ETH.

olga24912 (approved):

near/omni-tests/src/omni_token.rs
kisialiou: And where should it be checked? In the prover?

olga24912 (approved):

PR #149, opened 2024-11-29
fix: fix nonce encoding u128/u64

karim-en:


frolvanya (approved):
Should we fix CI or we can ignore warnings for this PR?

karim-en:
> Should we fix CI or we can ignore warnings for this PR?
yes, in a follow-up PR

PR #154, opened 2024-12-16
fix: fix wormhole address for Arbitrum

karim-en:


frolvanya (approved):

kiseln:
I believe this address is never going to change. Why do we need an update?

frolvanya:
@kiseln Because address initially was wrong. I believe @karim-en took a different address by mistake during the initialization

kiseln:
But it's not yet deployed on mainnet so update functionality seems redundant

olga24912 (approved):

PR #159, opened 2024-12-19
fix: normalize transferred tokens by decimals

karim-en:
### Problem:
On Solana, the max normal decimal of the SPL token is `9` and the `u64` is used for the transferred token amount, but on the Near side, the tokens usually have `24` decimals which prevents us from bridging this kind of tokens.

### Solition:
- Deploy tokens with decimals `18` on EVM and `9` Solana.
- Normolise and denormolise the transferred amount on the Near side.


evm/src/omni-bridge/contracts/BridgeTypes.sol
```diff
@@ -44,7 +44,8 @@ library BridgeTypes {
         string token,
         string name,
         string symbol,
-        uint8 decimals
+        uint8 decimals,
+        uint8 originDeciamls
```
frolvanya, line 48: should it be `originDecimals`?

karim-en: Fixed in 862fe1641646cd764427cc09d04121db5f4447b6

frolvanya: @karim-en Hmm, I see a fix, but you made a typo in new commit `Deciamls` -> `Decimals`

karim-en: 80602581078d32653ad061dbecfa35d4a552c5d2

frolvanya (approved):

kiseln:
Can we avoid usign near and do normalization/denormalization purely on destination chains?

For example, when we deploy on a destination chain we store original decimals and if it's higher than the threshold value we do transformations on each init/finalize

karim-en:
> Can we avoid usign near and do normalization/denormalization purely on destination chains?
> 
> For example, when we deploy on a destination chain we store original decimals and if it's higher than the threshold value we do transformations on each init/finalize

We can, but more changes will be needed + we will have a problem with the dust remainder which affect the total supply.
In this implementation the dust will be received by relayer with the fee.

near/omni-bridge/src/lib.rs
```diff
@@ -495,14 +504,19 @@ impl Contract {
             "Unknown factory"
         );
 
+        let decimals = self.token_decimals.get(&init_transfer.token);
```
olga24912, line 507: Maybe it makes sense to return an error if token_decimal is not found?

olga24912 (approved):

kiseln:
> > Can we avoid usign near and do normalization/denormalization purely on destination chains?
> > For example, when we deploy on a destination chain we store original decimals and if it's higher than the threshold value we do transformations on each init/finalize
> 
> We can, but more changes will be needed + we will have a problem with the dust remainder which affect the total supply. In this implementation the dust will be received by relayer with the fee.

I understand the dust issue but otherwise I see my suggested approach to be way simpler unless I'm missing something

near/omni-bridge/src/lib.rs
karim-en: But not all tokens have the decimal.
Do you propose also to insert the decimal in the `deploy_token_internal`? In this case, we should insert decimal with same value for `decimal` and `origin_decimal`

karim-en:
> > > Can we avoid usign near and do normalization/denormalization purely on destination chains?
> > > For example, when we deploy on a destination chain we store original decimals and if it's higher than the threshold value we do transformations on each init/finalize
> > 
> > 
> > We can, but more changes will be needed + we will have a problem with the dust remainder which affect the total supply. In this implementation the dust will be received by relayer with the fee.
> 
> I understand the dust issue but otherwise I see my suggested approach to be way simpler unless I'm missing something

Actually, I've tried to implement it like that, and found that is implementation become too complex, so I moved all this logic just to Near side. 
Anyway, feel free to try to implement POC, maybe you will be able to write it better than I did.

kisialiou:
@karim-en While the changes are approved, please, JIC pay attention to the fact that EVM tests failed and the Rust build failed as well.

near/omni-bridge/src/lib.rs
```diff
@@ -434,7 +446,14 @@ impl Contract {
     ) {
         if let Ok(signature) = call_result {
             if fee.is_zero() {
-                self.remove_transfer_message(message_payload.transfer_id);
+                let message = self.remove_transfer_message(message_payload.transfer_id);
+                let normalization_dust = message.amount.0 - message_payload.amount.0;
```
olga24912, line 450: What does this value represent, and why do we need it? We're subtracting the normalized value from the non-normalized one...

near/omni-bridge/src/lib.rs
olga24912: Yes, I think this approach will protect us from errors when we simply haven’t set the correct decimal yet.

near/omni-bridge/src/lib.rs
```diff
@@ -564,7 +588,11 @@ impl Contract {
         );
 
         let message = self.remove_transfer_message(fin_transfer.transfer_id);
-        let fee = message.amount.0 - fin_transfer.amount.0;
+        let de_normalized_amount = Self::de_normalize_amount(
+            fin_transfer.amount.0,
+            self.token_decimals.get(&message.token),
```
olga24912, line 593: We need tests to verify this pipeline. I suspect that message.token is the address of the original sender. So, in the case of a transfer from NEAR to Solana, it will be the address on NEAR. Consequently, the decimals will correspond to the NEAR address. However, to reconstruct how much was transferred, we would need the decimals from Solana.

olga24912 (changes requested):

near/omni-bridge/src/lib.rs
karim-en: The dust, or the difference between normalized and non-normalized values. This dust will be added to the fee on during the claim fee, but if the fee is zero then the dust will be stuck in the bridge contract.
The idea of storing this value was to be able to compare the total supply on different chains, but the benefit of this is debatable because the total supply is already hard to track if the the relayer decided not to claim the fee, so new tokens will not be minted.
So we can remove this logic and delegate this to the indexer.

olga24912 (approved):

evm/src/omni-bridge/contracts/BridgeTypes.sol
```diff
@@ -44,7 +44,8 @@ library BridgeTypes {
         string token,
         string name,
         string symbol,
-        uint8 decimals
+        uint8 decimals,
+        uint8 originDecimals
```
kiseln, line 48: Do we need `originDecimals`? Since we have `token` we can look it up on Near

solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/user/deploy_token.rs
```diff
@@ -86,6 +88,8 @@ impl<'info> DeployToken<'info> {
         let payload = DeployTokenResponse {
             token: metadata.token,
             solana_mint: self.mint.key(),
+            decimals: metadata.decimals,
+            origin_decimals,
```
kiseln, line 92: Same here

kiseln (approved):

evm/src/omni-bridge/contracts/BridgeTypes.sol
karim-en: It is a little bit complicated to retrieve the origin decimals due to async specific of the cross calls on Near, so by adding this we can reduce the number of callbacks on Near side

karim-en:
@race-of-sloths

race-of-sloths:
### ⏰️ PR is already merged
It's too late to include us now. Include us before the merge next time!

PR #160, opened 2024-12-20
Normalize to 18 decimals

kiseln:


evm/src/omni-bridge/contracts/OmniBridge.sol
```diff
@@ -283,6 +294,28 @@ contract OmniBridge is
         proxy.upgradeToAndCall(implementation, bytes(""));
     }
 
+    function denormalizeAmount(address tokenAddress, uint128 amount) internal view returns (uint128) {
+        uint8 originDecimals = tokenOriginDecimals[tokenAddress];
+        uint8 decimals = IERC20Metadata(tokenAddress).decimals();
+
+        if (originDecimals > decimals) {
+            return amount * uint128(10) ** (originDecimals - decimals);
```
frolvanya, line 302: We don't need this if statement. If `originDecimals == decimals`, then it would return `amount * 10^0` which is `amount`

evm/src/omni-bridge/contracts/OmniBridge.sol
```diff
@@ -283,6 +294,28 @@ contract OmniBridge is
...
+
+        return amount;
+    }
+
+    function normalizeAmount(address tokenAddress, uint128 amount) internal view returns (uint128) {
+        uint8 originDecimals = tokenOriginDecimals[tokenAddress];
+        uint8 decimals = IERC20Metadata(tokenAddress).decimals();
+
+        if (originDecimals > decimals) {
+            return amount / uint128(10) ** (originDecimals - decimals);
```
frolvanya, line 313: Same here

frolvanya (changes requested):

PR #166, opened 2024-12-25
Fix: fix fin transfer for EVM native token.

karim-en:


kiseln (approved):

evm/src/omni-bridge/contracts/OmniBridge.sol
```diff
@@ -194,7 +194,11 @@ contract OmniBridge is
             revert InvalidSignature();
         }
 
-        if (customMinters[payload.tokenAddress] != address(0)) {
+        if (payload.tokenAddress == address(0)) {
+            (bool success, ) = payload.recipient.call{value: payload.amount}("");
+            require(success, "ERR_FAILED_TO_SEND_ETHER");
```
kiseln, line 199: Let's use custom error

kiseln (changes requested):

olga24912 (approved):

evm/src/omni-bridge/contracts/OmniBridge.sol
karim-en: 9c45e46f4770fdfefa9d09c6a884187f06b87712

kiseln (approved):

Issue #165, opened 2024-12-24
[OmniBridge]: burn only (amount - fee) bridged tokens


olga24912:
In ft_on_transfer, the amount includes the fee. We burn the entire amount of tokens: 
https://github.com/Near-One/omni-bridge/blob/main/near/omni-bridge/src/lib.rs#L228

However, during `claim_fee`, we expect that the tokens for the fee are locked in the contract, not burned.

We need to burn `amount - fee` tokens instead.

olga24912:
https://github.com/Near-One/omni-bridge/pull/168

PR #168, opened 2024-12-25
Fix: don't burn the fee on ft_on_transfer

olga24912:
Issue: https://github.com/Near-One/omni-bridge/issues/165

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -225,7 +225,7 @@ impl FungibleTokenReceiver for Contract {
         if self.deployed_tokens.contains(&token_id) {
             ext_token::ext(token_id.clone())
                 .with_static_gas(BURN_TOKEN_GAS)
-                .burn((amount.0 - transfer_message.fee.fee.0).into());
+                .burn(amount);
         }
 
         env::log_str(&OmniBridgeEvent::InitTransferEvent { transfer_message }.to_log_string());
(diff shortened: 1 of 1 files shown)


frolvanya (approved):

near/omni-bridge/src/lib.rs
```diff
@@ -225,7 +225,7 @@ impl FungibleTokenReceiver for Contract {
         if self.deployed_tokens.contains(&token_id) {
             ext_token::ext(token_id.clone())
                 .with_static_gas(BURN_TOKEN_GAS)
-                .burn(amount);
+                .burn((amount.0 - transfer_message.fee.fee.0).into());
```
frolvanya, line 228: Nitpick: maybe we should use `parsed_msg`, since for `require_storage_balance` we use it as well and it removes one level of structure nesting?
```suggestion
                .burn((amount.0 - parsed_msg.fee.0).into());
```

olga24912: I think it doesn’t matter at all; it’s the same value. I chose transfer_message because we use it for the check above.

frolvanya: Okay, that makes sense

karim-en:
@olga24912 but what if the fee is updated?

karim-en:
> @olga24912 but what if the fee is updated?

I propose to burn everything and then mint on the claiming fee.

olga24912:
> > @olga24912 but what if the fee is updated?
> 
> I propose to burn everything and then mint on the claiming fee.

fixed: https://github.com/Near-One/omni-bridge/pull/168/commits/dbf920ec1c2430ebeb09c7437d9e2ab50333cdfc

karim-en (approved):

kiseln (approved):

kiseln:
> > @olga24912 but what if the fee is updated?
> 
> I propose to burn everything and then mint on the claiming fee.

Agree this is better. Plus, I think this is the only way for transfers between 2 non-near chains

PR #170, opened 2024-12-30
Add "Bridge New Token To Ethereum" workflow

kisialiou:
- Add contracts initialization.
- Add "Bridge New Token To Ethereum" workflow.


e2e-testing/Makefile
```diff
@@ -124,36 +142,187 @@ NEAR_BINARY_DIR := $(TESTING_ROOT)/near_artifacts
...
+		--token $$TOKEN_ID \
+		--near-signer $$SENDER_ACCOUNT_ID \
+		--near-private-key $$SENDER_PRIVATE_KEY \
+		--near-token-locker-id $$TOKEN_LOCKER_ID \
+		--config-file $(BRIDGE_SDK_CONFIG_FILE) > $@ && \
+	TX_HASH=$$(grep -o 'tx_hash="[^"]*"' $@ | cut -d'"' -f2) && \
+	echo "{\"tx_hash\": \"$$TX_HASH\"}" > $@
+
+# Step 2.2 (ETH PRIVATE KEY AND RPC ARE SUPPLIED IN SDK CONFIG)
+ethereum-deploy-token: $(EVM_DEPLOY_TOKEN_CALL_FILE)
```
olga24912, line 297: Issue: 
https://testnet.nearblocks.io/txns/9axKcT1FiEF2aw97kkVeYBZ4ymajSSCnEJ7NTxpVQDGU#execution

The previous transaction failed and needs to be redone. However, this doesn't happen automatically, and the failure of the transaction isn't being checked. As a result, the error occurs at a later stage than it should.

It seems necessary to add a command like "clear everything and restart the test from scratch.

olga24912 (changes requested):

e2e-testing/README.md
```diff
@@ -8,11 +8,20 @@
...
+- jq
 
 ## Using the Makefile
 
 The `Makefile` in this project is designed to automate the deployment and compilation processes for both EVM and NEAR environments. It provides a set of predefined rules that can be executed to perform specific tasks, such as compiling contracts, deploying them to various networks, and setting up necessary infrastructure.
 
+
+### Additional Requirements
+
+- **Private Key Requirement**: For Ethereum deployment, ensure that you add your `EVM_PRIVATE_KEY` to the `./evm/.env` file. This key is necessary for authenticating transactions on the Ethereum network. Also, update the `bridge-sdk-config.example.json` file with your `EVM_PRIVATE_KEY` and `ETH_RPC` and save it as `bridge-sdk-config.json`.
```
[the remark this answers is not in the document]
kisialiou: 1. We are using the same private key for all EVM chains, therefore it's called EVM_PRIVATE_KEY. In `bridge-sdk-config.example.json` I use the naming provided in the Bridge SDK and can't change it in this repo. I agree that it makes sense to change it to EVM_PRIVATE_KEY in Bridge SDK repo.
2. We are using the same private key for all EVM chains, including base and arbitrum.
3. ETH_RPC is required by Bridge SDK, not `evm_scripts`. I agree that configurations could be unified but I'd postpone it till later.
4. Agree.

e2e-testing/Makefile
kisialiou: Detecting the failure of the transaction (and moreover all of its receipts) on NEAR is not a straightforward task and is not supported in Bridge SDK. For now, I'd say that if the pipeline fails it's way easier to trace back the problematic place manually. Moreover, it tells us how far could the workflow go without detecting the failure.

However, it's a good point for the future. I've added it to the backlog.

olga24912 (approved):

PR #172, opened 2025-01-03
feat: implemented fee check

frolvanya:
> [!NOTE]
> To test this PR, https://github.com/Near-One/bridge-indexer-rs/issues/111 must be resolved first


omni-relayer/src/utils/fee.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let native_fee = response.native_token_fee.unwrap_or_default();
+    let transferred_fee = response.transferred_token_fee.unwrap_or_default();
+
+    match (native_fee, transferred_fee) {
+        (0, 0) => anyhow::bail!("No fee information found"),
+        (0, fee) if fee > 0 => Ok(provided_fee.fee.0 >= fee),
+        (native_fee, 0) if native_fee > 0 => Ok(provided_fee.native_fee.0 >= native_fee),
+        (native_fee, fee) => {
+            let conversion_rate = provided_fee.fee.0 as f64 / provided_fee.native_fee.0 as f64;
```
frolvanya, line 37: I don't like the idea of converting `u128` to `f64`, so let me know if there's a more elegant way to check "mixed" fee

omni-relayer/src/utils/fee.rs
```diff
@@ -0,0 +1,44 @@
...
+    let response = reqwest::get(&url)
+        .await?
+        .json::<TransferFeeResponse>()
+        .await?;
+
+    let native_fee = response.native_token_fee.unwrap_or_default();
+    let transferred_fee = response.transferred_token_fee.unwrap_or_default();
+
+    match (native_fee, transferred_fee) {
+        (0, 0) => anyhow::bail!("No fee information found"),
```
frolvanya, line 33: If both fee are `None` or `0` I mark it as an incorrect state of API. Is it okay?

omni-relayer/src/workers/near.rs
```diff
@@ -90,8 +90,23 @@ pub async fn sign_transfer(
...
-
-                        // TODO: Use existing API to check if fee is sufficient
+                        match utils::fee::is_fee_sufficient(
+                            &config,
+                            transfer_message.fee.clone(), 
+                            &transfer_message.sender,
+                            &transfer_message.recipient, 
+                            &transfer_message.token
+                        ).await {
+                            Ok(true) => {}
```
frolvanya, line 100: Is there a better way to do this? I can use `if let Err(_) | Ok(false) = ...`, but I'll not be able to see the state of error

omni-relayer/src/utils/fee.rs
olga24912: We should check: `(native_fee/fee) * provided_fee.fee + provided_fee.native_fee > native_fee` => 
`native_fee * provided_fee.fee + provided_fee.native_fee * fee > native_fee * fee`

We need to be cautious to avoid overflow during multiplication, and all operations must be performed safely. Intuitively, I think overflow is unlikely because fee is generally not a very large number. However, we still need to carefully consider this.

omni-relayer/src/workers/evm.rs
```diff
@@ -94,45 +98,97 @@ async fn handle_init_transfer_event(
...
+        };
+        let Ok(token) =
+            OmniAddress::new_from_evm_address(init_transfer_with_timestamp.chain_kind, token)
+        else {
+            warn!(
+                "Failed to convert token address to OmniAddress: {:?}",
+                init_log.inner.tokenAddress
+            );
+            return;
+        };
```
olga24912, line 164: Could you separate this logic for extracting OmniAddress into a dedicated function?

omni-relayer/src/workers/near.rs
olga24912: It is acceptable.

olga24912 (changes requested):

omni-relayer/src/utils/fee.rs
frolvanya: Oh, I messed up conversion rate, but I still can't do second option (native_fee * fee is dangerous, because 2^128 < 10^18 * 10^24)

frolvanya: Fixed in 30e9f72ffaa9d3b1d5215c6fb53de5ef0183aa10

omni-relayer/src/workers/evm.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/172/commits/a584a9bf7990a7236647bbf045bbe5ef111550fd

olga24912 (approved):

PR #176, opened 2025-01-06
Backward compatible address prefix for ETH network

karim-en:
The Rainbow Bridge deploys tokens in this format [<lower_case_hex>.factory.bridge.near](https://nearblocks.io/address/dac17f958d2ee523a2206206994597c13d831ec7.factory.bridge.near), after the migration we are going to use `factory.bridge.near` as deployer and the tokens for ethereum chain will be deployed in the same format.

### DIFF
--- a/near/omni-types/src/lib.rs
+++ b/near/omni-types/src/lib.rs
@@ -273,6 +273,7 @@ impl OmniAddress {
                     format!("sol-{hashed_address}")
                 }
             }
+            OmniAddress::Eth(address) => address.to_string(),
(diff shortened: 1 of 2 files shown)


frolvanya (approved):

olga24912 (changes requested):
Can you add a test to verify that the addresses are formatted correctly? I’m not entirely sure what address.to_string() will return—it might include a prefix like 0x

olga24912 (approved):

PR #177, opened 2025-01-10
feat: improve some utilities

frolvanya:


omni-relayer/src/config.rs
```diff
@@ -61,6 +61,7 @@ pub struct Evm {
     pub bridge_token_factory_address: Address,
     pub light_client: Option<AccountId>,
     pub block_processing_batch_size: u64,
+    pub expected_finalization_time: i64,
```
olga24912, line 64: Why i64?

olga24912 (approved):

omni-relayer/src/config.rs
frolvanya: Because I'm working with timestamps to check if we can start processing current event or not

timestamps are i64, so I want to keep everything in one type https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp

PR #183, opened 2025-01-12
feat: added mpc getter and updated near-plugins

frolvanya:


near/omni-bridge/src/lib.rs
```diff
@@ -394,6 +394,9 @@ impl Contract {
         fee: &Option<Fee>,
     ) -> Promise {
         let transfer_message = self.get_transfer_message(transfer_id);
+
+        require!(transfer_message.amount.0 > 0, "Invalid amount");
```
karim-en, line 398: This check should be done for the normalised amount

karim-en (changes requested):

near/omni-bridge/src/lib.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/183/commits/704963cb2a566390f6606d7380658b29d54d0435

kiseln:
What's the motivation for MPC getter?

frolvanya:
> What's the motivation for MPC getter?

To get required deposit for `sign` method from sdk
https://github.com/Near-One/bridge-sdk-rs/pull/66

frolvanya:
Also, one more thing to add, as you can see there's an issue with rustc version + lib versioning + wasm, since you need to optimize wasm after rustc v1.82.0 and we can't just use 1.81.0 or 1.79.0 since we upgraded near-plugins. I was able to compile wasm with 1.84.0 or run tests. During compilation I used `cargo-near`, so I want to rewrite our tests to use cargo-near instead and completely remove build.sh, because:
1) It'll get messier once we'll try to fix wasm with wasm-opt manually
2) Why reinvent the wheel?
3) It's a bad practice to store wasm files in a repo (especially if they're just used to check if build is the same)

Pros of `cargo-near`:
- ease of use
- we can build wasm directly from tests (aka `cargo nextest run` will build the latest wasm for us) https://github.com/NEAR-DevHub/neardevhub-contract/blob/main/tests/test_env/mod.rs#L30
- we could compile our contracts with ABI and this would allow source verifier to show the contract code to nearblocks.io (e.g https://nearblocks.io/address/devhub.near?tab=contract > contract code)
Cons:
- For now I don't see any. Let me know if you have something against it

karim-en (approved):

PR #184, opened 2025-01-13
feat: migration to cargo near

frolvanya:
# Usage Guide

## Compilation

There're two options to compile contract:
1) Using methods in `Makefile` from the root of repo
```bash
# I have added all packages, so general command looks like this: rust-build-<package-name>
$ make rust-build-evm-prover

# We can also build all contracts like this:
$ make rust-build-near
```

2) Using `cargo near` command directly:
```bash
$ cargo near build reproducible-wasm
# or we might need to specify `manifest-path` if the current pwd is not a contract (e.g pwd is ./near)
$ cargo near build reproducible-wasm --manifest-path ./omni-prover/evm-prover/Cargo.toml
```

> [!WARNING]
> If you have darwin-arm64 (aka MacBook with M chip) you would need to have rosetta installed (`softwareupdate --install-rosetta`). Otherwise, you might run in an issue during compilation with Docker

## Deployment

There're two options to deploy as well:
1) If we have already a wasm file that we want to deploy:
```bash
near contract deploy <account-id> use-file <path-to-wasm> with-init-call init json-args '{...}' prepaid-gas '100.0 Tgas' attached-deposit '0 NEAR' network-config <network> sign-with-keychain send
```

2) If we don't have a wasm we can either compile it using `cargo near build` or use `cargo near deploy`:
```bash 
cargo near deploy build-reproducible-wasm <account-id> without-init-call network-config testnet sign-with-keychain send
```
This command will build and deploy wasm for us

> [!NOTE]
> `cargo near` works in interactive mode as well just like `near-cli-rs`. Meaning you can call `cargo near` and process the rest in interactive mode by building command with a correct options


frolvanya:
rustc 1.84.0 + building wasm using cargo-near-build directly during testing works correctly: https://github.com/Near-One/omni-bridge/actions/runs/12738704841/job/35501410832?pr=184

karim-en (approved):

PR #209, opened 2025-01-27
Add pausability and admin methods

kiseln:
- Pausable admin can pause transfers
- Admin can unpause all or either direction
- Admin can also set new admin and pausable admin
- I avoided using upgrade authority for access control as it is not considered a good practice. In case we renounce upgradeability in the future we may still want to preserve admin control over the program.
- Changes were tested using squads (documentation to be added)


solana/bridge_token_factory/programs/bridge_token_factory/src/state/config.rs
```diff
@@ -22,5 +22,7 @@ pub struct Config {
     pub max_used_nonce: u64,
     pub derived_near_bridge_address: [u8; 64],
     pub bumps: ConfigBumps,
-    pub padding: [u8; 100],
+    pub paused: u8,
+    pub pausable_admin: Pubkey,
+    pub padding: [u8; 67],
```
karim-en, line 27: Do we need a migration for this, or is decreasing the padding enough?

karim-en (commented):
Please fix the conflicts and CI errors

solana/bridge_token_factory/programs/bridge_token_factory/src/state/config.rs
kiseln: It's enough

karim-en (approved):

PR #214, opened 2025-01-29
feat: retry sign_transfer

frolvanya:
Check if all receipts are successful and if it's not the case, retry signing process


karim-en (commented):
@frolvanya please merge with the main branch

frolvanya:
> @frolvanya please merge with the main branch

I can't do it, base branch change requires another approval. Also, I needed to fix a merge conflict with main. So I need your approval

omni-relayer/src/workers/solana.rs
```diff
@@ -291,6 +291,19 @@ async fn handle_init_transfer_event(
         Ok(actions) => actions,
         Err(err) => {
             warn!("{}", err);
+            utils::redis::remove_event(
+                &mut redis_connection,
+                utils::redis::SOLANA_INIT_TRANSFER_EVENTS,
+                &key,
+            )
+            .await;
+            utils::redis::add_event(
```
karim-en, line 300: There is a chance that the `remove_event` and the `add_event` not.
So we need to use redis transaction here, or just swap these two actions.

omni-relayer/src/workers/near.rs
```diff
@@ -19,6 +20,8 @@ use omni_types::{
 
 use crate::{config, utils};
 
+const NONCE_ALREADY_USED_ERROR: i64 = -32002;
```
karim-en, line 23: Where did this number come from? And how to distinguish between `NONCE_ALREADY_USED_ERROR` and lake of balance or other contract error?

karim-en (changes requested):

omni-relayer/src/workers/solana.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/214/commits/a2a598d2329e92f7b8f84b8fffb18e72261623bf

omni-relayer/src/workers/near.rs
frolvanya: image
That's a screenshot from #216. We can add more codes, but for now we need only this one. In case of lack of balance we still need to retry (until someone will top up relayer balance), we don't need to wipe record from db in this case

karim-en: You can't use this error code, because error code `-32002` is general so it is possible to get it due to other problems. Please add a check for the contract error `NonceAlreadyUsed`

karim-en: https://ethereum-json-rpc.com/errors

karim-en: @frolvanya check this 
https://docs.rs/ethers/latest/ethers/contract/enum.ContractError.html#method.decode_revert

frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/214/commits/04554fd3ea340cf12d963413b475c6d2e6fe972d

frolvanya: > @frolvanya check this https://docs.rs/ethers/latest/ethers/contract/enum.ContractError.html#method.decode_revert
> 
> https://github.com/gakonst/ethers-rs/blob/6e2ff0ef8af8c0ee3c21b7e1960f8c025bcd5588/ethers-contract/tests/it/contract.rs#L704

Not sure how this would help me. I'm expecting error from Solana, these are docs related to Eth

frolvanya: @karim-en I took 6003 (0x1773) from here:
image

This error should be tied only to `NonceAlreadyUsed`

frolvanya: @karim-en https://github.com/Near-One/omni-bridge/blob/679d48018ee2272408e4207212b2133e25d8ebf5/solana/bridge_token_factory/programs/bridge_token_factory/src/error.rs#L12

6000 (offset) + 3 (`NonceAlreadyUsed` position in enum)

karim-en (approved):

PR #221, opened 2025-01-31
fix: restart ws connection to rpc when it was dropped

### LINKED ISSUE #220: [OmniRelayer]: restart web socket connection for indexers if it was dropped
Recently relayer died because Eth indexer's ws connection stopped working:

Image

frolvanya:
image

In the worst case scenario when there's no network connection relayer will try to reconnect to rpc providers every 5 seconds instead of dying silently

closes #220

### DIFF
--- a/omni-relayer/src/startup/evm.rs
+++ b/omni-relayer/src/startup/evm.rs
@@ -53,22 +53,31 @@ pub async fn start_indexer(
         "Failed to parse {chain_kind:?} rpc provider as url",
     ))?);
 
-    let ws_provider = ProviderBuilder::new()
-        .on_ws(WsConnect::new(rpc_ws_url))
-        .await
-        .context(format!("Failed to initialize {chain_kind:?} WS provider"))?;
-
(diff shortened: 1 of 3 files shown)


omni-relayer/src/startup/evm.rs
```diff
@@ -110,19 +119,50 @@ pub async fn start_indexer(
...
+            .on_ws(WsConnect::new(&rpc_ws_url))
+            .await
+        {
+            Ok(provider) => provider,
+            Err(err) => {
+                error!(
+                    "{chain_kind:?} WebSocket connection failed: {}, retrying...",
+                    err
+                );
+                tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
karim-en, line 133: This can be refactored using macros.
See this example 
https://github.com/Near-One/eth2-on-near-client-validator/blob/48fe316da190573f523e37700c35ad557b38db67/src/eth2_validator.rs#L15

karim-en (changes requested):

karim-en (approved):

PR #224, opened 2025-02-03
feat: added message field for solana transfers

frolvanya:
This PR should be merged only when new version of contract on solana will be deployed


frolvanya:
@karim-en yes, this must be added here:
https://github.com/Near-One/bridge-indexer-rs/blob/b49669a5a7c17f72e68ee4f125ea7375d0c8559a/solana-indexer/src/parser.rs#L28

karim-en:
> @karim-en yes, `message` field must be added here: https://github.com/Near-One/bridge-indexer-rs/blob/b49669a5a7c17f72e68ee4f125ea7375d0c8559a/solana-indexer/src/parser.rs#L28

But we should be able to parse both versions of the payloads, can we add a fallback to try to deserialize with the old format if the new one fails ?

frolvanya:
@karim-en hmm, we can make `message` field optional. But why do we need to be able to parse both versions? All older tx are processed

karim-en:
> @karim-en hmm, we can make `message` field optional. But why do we need to be able to parse both versions? All older tx are processed

It is ok for the relayer, but the indexer should be able to process both kind of messages.

karim-en (approved):

PR #227, opened 2025-02-04
fix: restart ws connection to rpc when it was dropped and sync lost blocks

frolvanya:
We found an issue when connection can be dropped right when new event should be found, so we need to fetch everything from the last_processed_block
image
image

### DIFF
--- a/omni-relayer/src/startup/evm.rs
+++ b/omni-relayer/src/startup/evm.rs
@@ -32,7 +32,7 @@ pub async fn start_indexer(
     config: config::Config,
     redis_client: redis::Client,
     chain_kind: ChainKind,
-    start_block: Option<u64>,
+    mut start_block: Option<u64>,
 ) -> Result<()> {
     let mut redis_connection = redis_client.get_multiplexed_tokio_connection().await?;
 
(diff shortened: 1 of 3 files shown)


omni-relayer/src/startup/evm.rs
```diff
@@ -49,25 +49,97 @@ pub async fn start_indexer(
...
+                &mut redis_connection,
+                &http_provider,
+                log,
+                expected_finalization_time,
+            )
+            .await;
+        }
 
+        error!("{chain_kind:?} WebSocket stream closed unexpectedly, reconnecting...");
+        start_block = None;
```
frolvanya, line 113: This is done to make sure that first start uses `start_block` (if it was provided as an argument) and then we need to switch back to using `last_processed_block`. Let me know if this should be refactored using better code practices

Also, same logic can be found in solana startup files

karim-en: Usually, it is done in this way:
```
let mut last_processed_block = start_block.unwrap_or(redis.get_last_processed_block());
loop {
    lake::start(last_processed_block, &mut last_processed_block)
}
```
But you implementation is ok for now

karim-en (approved):

PR #245, opened 2025-02-11
Automate Solana deployment with Make

kiseln:


solana/bridge_token_factory/programs/bridge_token_factory/build.rs
```diff
@@ -4,7 +4,7 @@ fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
     let program_id =
-        env::var("PROGRAM_ID").unwrap_or("dahPEoZGXfyV58JqqH85okdHmpN8U2q8owgPUXSCPxe".into());
+        env::var("PROGRAM_ID").unwrap();
```
karim-en, line 7: @diegofigs we want to remove the default  `PROGRAM_ID` to avoid mistakes during the building and deployment but we break the clippy. Is it possible to set this env variable for the backstage flow, eg adding it to the `.catalog-info.yam`? 
CC @kiseln

diegofigs: @karim-en there's a workaround: the workflow calls a Makefile's `env` script, in there you could define the `PROGRAM_ID`

Makefile
```diff
@@ -58,8 +58,22 @@ rust-build-mock-token:
 
 rust-build-near: rust-build-omni-bridge rust-build-omni-token rust-build-token-deployer rust-build-omni-prover rust-build-evm-prover rust-build-wormhole-omni-prover-proxy rust-build-mock-prover rust-build-mock-token
 
-rust-build-solana:
-	cd solana/bridge_token_factory && RUSTUP_TOOLCHAIN="nightly-2024-11-19" anchor build --verifiable
+solana-generate-program-id:
```
karim-en, line 61: @kiseln  the `rust-build-solana` was used in this [workflow](https://github.com/Near-One/omni-bridge/blob/533f2fde3ee59c0d4c2bbde2e8797fda4b0293ce/.github/workflows/update-contracts.yaml#L33), can you please fix it.

kiseln: 9f38bfda2255921c977dd6b60aabce127d72da08

karim-en: I think this will not work because there is no key file.

kiseln: All checks have passed. I don't think we need a key file for building

karim-en: It passed because the build release is trigger only on creating a tags

Issue #251, opened 2025-02-12
[BUG] Large native NEAR withdrawals will be stuck forever


mitinarseny:
Large amounts of native NEAR will always fail to finalize and, thus, remain stuck in `omni-bridge` contract forever:

https://github.com/Near-One/omni-bridge/blob/707df6f4d5dfc1c2a92a0bbe31bb3a3d1e922bce/near/omni-bridge/src/lib.rs#L1042-L1046

An example situation when this might happen would be if `omni-bridge` has only 100 native NEAR balance, while relayer tries to finalize 1000 native NEAR withdrawal, even if the contract owns 1000+ wNEAR.

The reason for that is sharded Near architecture: when Near VM schedules a native transfer `Promise` to increase balance of recipient, it also subtracts the `amount` from `current_account_id` **during execution of current receipt**, but not when scheduled promise is executed. If there is not enough native NEAR balance on the contract **at the time of `Promise` creation**, then current receipt will fail with `ExecutionError("Exceeded the account balance.")`. 

To mitigate this issue, an additional `Promise` should be [scheduled as a callback](https://github.com/near/intents/blob/b59fd6a016d1a064f5450305c5a92744598f88a9/defuse/src/contract/intents/state.rs#L171-L177) to `near_withdraw`, and the actual transfer `Promise` should only be [created in this callback](https://github.com/near/intents/blob/b59fd6a016d1a064f5450305c5a92744598f88a9/defuse/src/contract/tokens/nep141/native.rs#L12-L20):
```rust
// first withdraw
.near_withdraw(U128(withdraw.amount.as_yoctonear()))
.then(
    // do_native_withdraw only after unwrapping NEAR
    Contract::ext(CURRENT_ACCOUNT_ID.clone())
        .with_static_gas(Contract::DO_NATIVE_WITHDRAW_GAS)
        .do_native_withdraw(withdraw),
)

// ...
#[private]
pub fn do_native_withdraw(&mut self, withdraw: NativeWithdraw) -> Promise {
    require!(
        matches!(env::promise_result(0), PromiseResult::Successful(data) if data.is_empty()),
        "near_withdraw failed",
    );

    Promise::new(withdraw.receiver_id).transfer(withdraw.amount)
}

```

[...]

PR #259, opened 2025-02-14
Fix: fix unwrap `wNear`

karim-en:
https://github.com/Near-One/omni-bridge/issues/251


frolvanya (approved):

frolvanya:
@karim-en since new functionality was added to omni-bridge, we need to update version in toml and create new release

kiseln (approved):

PR #263, opened 2025-02-16
Solana: Add update metadata method and an updater role

kiseln:
- `update_metadata` method for updating metadata uri
- `metadata_admin` role that can update metadata

### DIFF
--- a/solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/initialize.rs
+++ b/solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/initialize.rs
@@ -94,6 +94,7 @@ impl Initialize<'_> {
         &mut self,
         admin: Pubkey,
         pausable_admin: Pubkey,
+        metadata_admin: Pubkey,
         derived_near_bridge_address: [u8; 64],
         config_bump: u8,
         authority_bump: u8,
@@ -118,7 +119,8 @@ impl Initialize<'_> {
             },
             paused: 0,
             pausable_admin,
-            padding: [0; 67],
+            metadata_admin,
+            padding: [0; 35],
         });
 
         let rent = Rent::get()?;
--- a/solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
+++ b/solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
@@ -1,0 +1,88 @@
+use anchor_lang::prelude::*;
+use anchor_lang::{
+    prelude::{Interface, SystemAccount},
+    Accounts, Key,
+};
+use anchor_spl::metadata::mpl_token_metadata::types::DataV2;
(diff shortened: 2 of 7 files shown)


solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
```diff
@@ -0,0 +1,88 @@
...
+            self.token_program.to_account_info(),
+            cpi_accounts,
+            signer_seeds,
+        );
+
+        update_metadata_accounts_v2(
+            cpi_ctx,
+            None,
+            Some(DataV2 {
+                name: self.metadata.name.clone(),
```
karim-en, line 74: On Near we also allow updating the name and symbol, I think this could be useful here, but pass them as Option<>

solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/initialize.rs
```diff
@@ -118,7 +119,8 @@ impl Initialize<'_> {
             },
             paused: 0,
             pausable_admin,
-            padding: [0; 67],
+            metadata_admin,
```
karim-en, line 122: Since the role management is more complex for Solana than Near, then we can later move all the management to Near, so we can sign messages with specific actions.

solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
```diff
@@ -0,0 +1,88 @@
+use anchor_lang::prelude::*;
+use anchor_lang::{
+    prelude::{Interface, SystemAccount},
+    Accounts, Key,
+};
+use anchor_spl::metadata::mpl_token_metadata::types::DataV2;
+use anchor_spl::metadata::{
+    update_metadata_accounts_v2, Metadata, MetadataAccount, UpdateMetadataAccountsV2,
+};
+use anchor_spl::{metadata::ID as MetaplexID, token::Mint, token_interface::TokenInterface};
```
karim-en, line 10: Why `MetaplexID`?

karim-en (changes requested):

solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
kiseln: It is an Id of Metaplex metadata program

solana/bridge_token_factory/programs/bridge_token_factory/src/instructions/admin/update_metadata.rs
kiseln: 1ca0537adfc786638f2c84349d42e189ece4686c

karim-en (approved):

PR #265, opened 2025-02-17
feat: pass the fast relayer over `FastFinTransferMsg`

karim-en:
The fast relayer could use external protocols to hold and swap liquidity so that the sender_id will be the protocol contract instead of the real relayer ID. 
In the PR, the relayer id should be passed in the `FastFinTransferMsg`.

### DIFF
--- a/near/omni-types/src/lib.rs
+++ b/near/omni-types/src/lib.rs
@@ -393,6 +393,7 @@ pub struct FastFinTransferMsg {
     pub fee: Fee,
     pub msg: String,
     pub storage_deposit_amount: Option<u128>,
+    pub relayer: FastRelayer,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


near/omni-types/src/lib.rs
```diff
@@ -556,7 +550,8 @@ impl FastTransfer {
     }
 }
 
-#[derive(BorshDeserialize, BorshSerialize, Debug, Clone)]
+#[near(serializers=[borsh, json])]
```
kiseln, line 553: Is json needed here?

near/omni-types/src/lib.rs
```diff
@@ -393,6 +393,7 @@ pub struct FastFinTransferMsg {
     pub fee: Fee,
     pub msg: String,
     pub storage_deposit_amount: Option<u128>,
+    pub relayer: AccountId,
```
kiseln, line 396: Perhaps we can make it optional and take sender_id otherwise?

kiseln (approved):

near/omni-types/src/lib.rs
karim-en: For now no, but if we are going to add an getter then yes

near/omni-types/src/lib.rs
karim-en: Yeah, I considered doing that too, but I realized that the risk of failing to pass this is greater than the inconvenience of adding it.

PR #272, opened 2025-02-20
Add the ability to receive ETH in Omni Bridge.

olga24912:
The ability to receive Ether is necessary to facilitate the migration from Rainbow Bridge. Ether must be locked in this contract to enable unlock operations, so we need to transfer Ether to this contract from the EthCustodian.

### DIFF
--- a/evm/hardhat.config.ts
+++ b/evm/hardhat.config.ts
@@ -107,6 +107,18 @@ task("deploy-bridge-token-factory", "Deploys the OmniBridge contract")
     )
   })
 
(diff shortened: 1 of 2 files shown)


karim-en (approved):

kiseln:
Alternatively could add a specialized payable migration method. Could be a little safer in case we don't want to receive ETH normally

olga24912:
> Alternatively could add a specialized payable migration method. Could be a little safer in case we don't want to receive ETH normally. I'm okay to merge as it is though

Or we can restrict access to this function to only the admin.

karim-en (approved):

PR #273, opened 2025-02-20
Update `near-lake-framework`

karim-en:


frolvanya (approved):
Please, use `cargo update` to update lock file as well

frolvanya (approved):

PR #280, opened 2025-02-26
feat: add native fee restriction

### LINKED ISSUE #277: [OmniBridge] Add blacklist for native fee payer
The Aurora engine can't attach Near as a native fee because the native token on the Aurora EVM is ETH.
We should rethink how the Aurora EVM should be integrated with OmniBridge, but for now are going to pay the storage in advance for init transfers and restrict paying in native fee to avoid withdrawing the deposited balance.

1- Add a blacklist map where some sender_id aren't allowed to attach native Near
2- I recommend implementing it in a generic way to avoid adding other maps later for whitelisting or other actions, so the map could looks like this `near_sdk::map<OmniAddress, std::vec<EnumRule>>`

r-near:
## Problem

Aurora cannot attach NEAR tokens to cover storage costs and fees. To address this limitation:

1. Aurora could pre-deposit large amounts of NEAR (e.g., 100 NEAR) to cover storage costs for these transfers
2. The contract currently manages two types of balances:
   - Storage deposit balance (tracked in `accounts_balances`)
   - Native fee handling for cross-chain transfers

The issue exists in how native fees are processed during `ft_on_transfer`:
```rust
let transfer_message = TransferMessage {
    // ...other fields...
    fee: Fee {
        fee: parsed_msg.fee,
        native_fee: parsed_msg.native_token_fee,
    },
    // ...other fields...
};

// Add the native_token_fee to required_storage_balance
required_storage_balance = required_storage_balance
    .saturating_add(NearToken::from_yoctonear(parsed_msg.native_token_fee.0));

// Deduct from sender's available balance
self.update_storage_balance(
    sender_id,
    required_storage_balance,
    NearToken::from_yoctonear(0),
);
```

A malicious user could exploit this by:
1. Initiating transfers via these external chains with extremely high `native_token_fee` values
2. The contract would deduct these fees from the pre-deposited funds
3. This could quickly drain all pre-deposited funds (100 NEAR), making the bridge unusable for legitimate users

## Solution

This PR implements a role-based restriction mechanism using the existing ACL system from near-plugins:

1. Added `NativeFeeRestricted` role to the `Role` enum
2. Modified `ft_on_transfer` to check if the sender has this role and reject transfers with non-zero native fees
3. Added comprehensive tests to verify the functionality

## Implementation Details

- The implementation leverages the existing ACL system rather than creating a separate blacklist
- Added `NativeFeeRestricted` role to the `Role` enum:
- Modified `ft_on_transfer` to check for restricted accounts:
- Bridge Operator accounts can be granted the `NativeFeeRestricted` role to prevent them from using native fees
- This prevents potential abuse where accounts could set excessive native fees and drain pre-deposited funds
- Tests confirm that:
  - Accounts with the role cannot set native fees
  - Accounts without the role can set native fees normally
  - Role management works correctly (granting/revoking)

## Technical Considerations

1. **Alternative Solutions Considered**:
   - Creating a separate native fee balance tracking system was deemed too complex and would require extensive changes
   - Implementing a dedicated blacklist data structure would add redundancy since we already have the ACL system

2. **Security Benefits**:
   - Prevents potential denial of service by protecting pre-deposited funds
   - Uses the battle-tested ACL system which has undergone security audits
   - Role permissions ensure only authorized accounts (DAO) can manage restrictions

3. **Administrative Process**:
   - DAO members can grant the role: `acl_grant_role("NativeFeeRestricted", "bridge-account.near")`
   - Bridge operators will pre-deposit their storage funds and be restricted from using native fees
   - Normal users remain unaffected by this change

This provides a secure way to prevent the vulnerability while maintaining compatibility with the existing system. It's a more elegant solution than creating a separate blacklist structure, as it integrates with the existing permission model.

Closes #277

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -199,7 +199,7 @@ impl FungibleTokenReceiver for Contract {
     ) -> PromiseOrValue<U128> {
         let parsed_msg: InitTransferMsg = serde_json::from_str(&msg).sdk_expect("ERR_PARSE_MSG");
         let token_id = env::predecessor_account_id();
-        
+
         // User has to pay for storage and we can't trust sender_id.
         let signer_id = env::signer_account_id();
 
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -204,10 +204,8 @@ impl FungibleTokenReceiver for Contract {
         let signer_id = env::signer_account_id();
 
         // Avoid extra storage read by verifying native fee before checking the role
-        if parsed_msg.native_token_fee.0 > 0 {
-            if self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone()) {
-                env::panic_str("ERR_ACCOUNT_RESTRICTED_FROM_USING_NATIVE_FEE");
-            }
+        if parsed_msg.native_token_fee.0 > 0 && self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone()) {
+            env::panic_str("ERR_ACCOUNT_RESTRICTED_FROM_USING_NATIVE_FEE");
         }
 
         require!(
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -203,10 +203,11 @@ impl FungibleTokenReceiver for Contract {
         // User has to pay for storage and we can't trust sender_id.
         let signer_id = env::signer_account_id();
 
-        if self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone())
-            && parsed_msg.native_token_fee.0 > 0
-        {
-            env::panic_str("ERR_ACCOUNT_RESTRICTED_FROM_USING_NATIVE_FEE");
+        // Avoid extra storage read by verifying native fee before checking the role
+        if parsed_msg.native_token_fee.0 > 0 {
+            if self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone()) {
+                env::panic_str("ERR_ACCOUNT_RESTRICTED_FROM_USING_NATIVE_FEE");
+            }
         }
 
         require!(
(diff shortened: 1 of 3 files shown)


near/omni-bridge/src/lib.rs
```diff
@@ -198,6 +199,13 @@ impl FungibleTokenReceiver for Contract {
     ) -> PromiseOrValue<U128> {
         let parsed_msg: InitTransferMsg = serde_json::from_str(&msg).sdk_expect("ERR_PARSE_MSG");
         let token_id = env::predecessor_account_id();
+
+        if self.acl_has_role(Role::NativeFeeRestricted.into(), sender_id.clone())
```
karim-en, line 203: Please merge with main and change the `sender_id` to `signer_id` due to this changes https://github.com/Near-One/omni-bridge/pull/284

karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -198,6 +199,13 @@ impl FungibleTokenReceiver for Contract {
     ) -> PromiseOrValue<U128> {
         let parsed_msg: InitTransferMsg = serde_json::from_str(&msg).sdk_expect("ERR_PARSE_MSG");
         let token_id = env::predecessor_account_id();
+
+        if self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone())
```
karim-en, line 203: The signer_id isn't defined here, you need to move these lines up
```
        // User has to pay for storage and we can't trust sender_id.
        let signer_id = env::signer_account_id();
```

karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -198,6 +199,16 @@ impl FungibleTokenReceiver for Contract {
     ) -> PromiseOrValue<U128> {
         let parsed_msg: InitTransferMsg = serde_json::from_str(&msg).sdk_expect("ERR_PARSE_MSG");
         let token_id = env::predecessor_account_id();
+
+        // User has to pay for storage and we can't trust sender_id.
+        let signer_id = env::signer_account_id();
+
+        if self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone())
```
karim-en, line 206: Let's optimizing it a little bit and avoid extra storage read by verifying the native fee first

karim-en (changes requested):

karim-en (approved):

PR #282, opened 2025-02-26
fix: check fee

frolvanya:


karim-en:
@frolvanya can you please bump the version, so we can release it

PR #286, opened 2025-03-04
feat: bridge-indexer integration

frolvanya:


frolvanya:
Since now we have a dependency from `bridge-indexer-types` we need to fix our CI, since this crate is in private repo. Maybe this can be done by adding some access keys to GitHub secrets

omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -0,0 +1,469 @@
...
+}
+
+async fn watch_omni_events_collection(
+    collection: &Collection<OmniEvent>,
+    mut redis_connection: redis::aio::MultiplexedConnection,
+    config: &config::Config,
+) {
+    let resume_token: Option<ResumeToken> = utils::redis::get_last_processed::<&str, String>(
+        &mut redis_connection,
+        utils::redis::MONGODB_OMNI_EVENTS_RT,
```
olga24912, line 366: Right now I have the impression that there are two approaches to the indexer, and the relayer can be switched back and forth between them. What will happen when we switch from the old approach to the new one? Will we be processing events from the very beginning?

omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -0,0 +1,469 @@
...
+
+                                async move {
+                                    if let Err(err) = handle_transaction_event(
+                                        redis_connection,
+                                        config,
+                                        event.transaction_id,
+                                        event.origin,
+                                        transaction_event,
+                                    )
+                                    .await
```
olga24912, line 392: Let's say we lose the connection to Redis and nothing gets written there at all. Then we would just skip an arbitrary number of transactions until the connection is restored?

omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -0,0 +1,469 @@
...
+            };
+
+            let Some(expected_finalization_time) = (match chain_kind {
+                ChainKind::Eth => config.eth.map(|eth| eth.expected_finalization_time),
+                ChainKind::Base => config.base.map(|base| base.expected_finalization_time),
+                ChainKind::Arb => config.arb.map(|arb| arb.expected_finalization_time),
+                _ => None,
+            }) else {
+                anyhow::bail!(
+                    "Failed to get expected_finalization_time, since config for {:?} is not set",
```
olga24912, line 127: This is an issue in our config, not in the transaction itself, so it would be strange if we skipped this transaction and never came back to it.

omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -0,0 +1,469 @@
...
+                    chain_kind
+                );
+            };
```
olga24912, line 185: Copying the previous case. Better to move it into a separate function.

omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -0,0 +1,469 @@
...
```
olga24912, line 320: This part is already repeated for the third time. L283-L320

olga24912 (changes requested):

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: If we start relayer with native indexers, then we will start from last processed block stored in redis
If we start relayer with bridge indexer, then we will use resume token, if no resume token is found, then we will process only new events that will come after the point when relayer was started

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: yes, but I don't see a situation where this can happen. We have 5 or 10 retries for each write operation to redis

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: Should we panic on this case then?

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/286/commits/1377033a4c93e15977c2b342e97b9e0ce130d766

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/286/commits/1377033a4c93e15977c2b342e97b9e0ce130d766

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/286/commits/7b0e61411f3f28d2c61c16015381bb04757d1d3b

Let me know if this is not a desired fix

frolvanya:
Well, looks like copilot is still not able to check if modules/files exists before writing a comment 😅

omni-relayer/src/startup/bridge_indexer.rs
olga24912: How can we switch from one indexer to another and make sure we haven't lost any events?

omni-relayer/src/startup/bridge_indexer.rs
olga24912: Usually, Redis works quite stably. However, if Redis is completely down and, say, needs to be restarted, then several retries in a row won't help. 

Yes, we can create a separate issue and then think about how to solve this problem.

olga24912 (approved):

omni-relayer/src/startup/bridge_indexer.rs
frolvanya: just start indexer with native relayers (comment out `mongodb_uri` or `db_name` fields from `bridge_indexer` section in `config.toml`) with 5 (for every chain) `--start-block` options and pass data from checkpoints there:
image

karim-en (approved):

PR #288, opened 2025-03-05
fix: retry when encounter timeout on near-rpc

frolvanya:


omni-relayer/src/workers/mod.rs
```diff
@@ -1193,7 +1217,9 @@ async fn process_evm_fin_transfer_event(
 
             if let BridgeSdkError::NearRpcError(near_rpc_error) = err {
                 match near_rpc_error {
-                    NearRpcError::NonceError | NearRpcError::FinalizationError => {
+                    NearRpcError::NonceError
+                    | NearRpcError::FinalizationError
+                    | NearRpcError::RpcTransactionError(JsonRpcError::TransportError(_)) => {
```
karim-en, line 1222: It will be good to reduce the code duplication in a follow PR

karim-en (approved):

PR #290, opened 2025-03-07
Refactor ERC20-NEAR Bridge Checker to TypeScript + Zorsh

r-near:
This PR refactors the ERC20 Ethereum to NEAR Bridge transaction finalization verifier with these technical improvements:

## Changes

- **TypeScript Migration**: Full rewrite from JavaScript to TypeScript
- **[Zorsh](https://github.com/r-near/zorsh) Integration**: I wrote a library for type-safe Borsh serialization - switched to that.
- **Type Safety**: Added proper typings for EVM proofs and all data structures
- **Environment Validation**: Added Envalid for type-safe environment variable validation
- **Error Handling**: Implemented robust error catching for each transaction
- **UX Improvements**: Added console coloring and better progress reporting
- **Dev Setup**: Updated to modern pnpm/tsx execution environment

## Technical Notes

- Zorsh provides a strongly-typed schema-based approach to Borsh, which simplifies the serialization of EVM proofs
- Type-safe serialization eliminates several potential bugs in the original implementation
- Proper error isolation prevents individual transaction failures from crashing the entire process
- The EVMProofSchema now has correct BigInt handling for log_index and receipt_index

## Notes

- Run with `pnpm tsx index.ts` instead of the previous `node index.js`
- Environment variables schema is now validated at runtime
- Added TypeScript types for improved IDE support


migration/checkERC20EthToNearTransfersFinalization/index.ts
```diff
@@ -0,0 +1,203 @@
...
+        const isFinalized = await isUsedProof(proofLight);
+
+        if (!isFinalized) {
+          notFinalizedCount += 1;
+          console.log(chalk.yellow(`Transaction ${event.transactionHash} is NOT finalized`));
+        } else {
+          console.log(chalk.green(`Transaction ${event.transactionHash} is finalized`));
+        }
+      } catch (error) {
+        console.error(chalk.red(`Error processing transaction ${event.transactionHash}:`), error);
```
olga24912, line 173: If an error occurs here, the message "All transactions are finalized" will still appear at the end.

migration/checkERC20EthToNearTransfersFinalization/index.ts
```diff
@@ -0,0 +1,203 @@
...
+    
+    // Process each event
+    for (const event of events) {
+      console.log(chalk.cyan(`\nProcessing transaction ${event.transactionHash}`));
+      
+      try {
+        // Get block data
+        const block = await provider.send(
+          'eth_getBlockByNumber',
+          [ethers.toBeHex(event.blockNumber), false]
```
olga24912, line 133: It returns error: 
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'UNKNOWN_ERROR',
  error: {
    code: -32602,
    message: 'invalid argument 0: hex number with leading zero digits'
  },
  payload: {
    method: 'eth_getBlockByNumber',
    params: [ '0x0151885b', false ],
    id: 5,
    jsonrpc: '2.0'
  },
  shortMessage: 'could not coalesce error'
}


Fix something like that:
```
        const hexBlockNumber = '0x' + BigInt(event.blockNumber).toString(16);

        const block = await provider.send(
          'eth_getBlockByNumber',
          [hexBlockNumber, false]
        );
```

olga24912 (changes requested):

migration/checkERC20EthToNearTransfersFinalization/index.ts
karim-en: cc @r-near

migration/checkERC20EthToNearTransfersFinalization/README.md
```diff
@@ -2,20 +2,58 @@
...
+   ```
+
+3. Create an `.env` file in the project root with the following variables:
+   ```
+   INFURA_API_KEY=your_infura_api_key
+   NETWORK_ETH=goerli_or_mainnet
+   ERC20_LOCKER=erc20_locker_contract_address
+   NETWORK_NEAR=testnet_or_mainnet
+   BRIDGE_TOKEN_FACTORY_ACCOUNT_ID=bridge_token_factory_account
+   NEAR_ACCOUNT_ID=your_near_account.near
```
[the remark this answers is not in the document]
r-near: I think we could probably use that "universal" env in the main `migration` directory.

migration/checkERC20EthToNearTransfersFinalization/index.ts
r-near: Ah it looks like we need to use `toQuantity` instead of `toBeHex`:
```javascript
/**
 *  Returns a [[HexString]] for %%value%% safe to use as a //Quantity//.
 *
 *  A //Quantity// does not have and leading 0 values unless the value is
 *  the literal value `0x0`. This is most commonly used for JSSON-RPC
 *  numeric values.
 */
 ```

File: migration/checkERC20EthToNearTransfersFinalization/index.ts (hunk 1 above)
Comment (r-near) [reply migration/checkERC20EthToNearTransfersFinalization/index.ts]: Now will exit early.

File: migration/checkERC20EthToNearTransfersFinalization/index.ts [hunk 4]
```diff
@@ -0,0 +1,234 @@
...
+// Load environment variables
+dotenv.config();
+
+// Validate environment variables with envalid
+const ENV = cleanEnv(process.env, {
+	ERC20_LOCKER: str({ desc: "Ethereum address of the ERC20 Locker contract" }),
+	NETWORK_ETH: str({ choices: ["mainnet", "goerli", "sepolia"] }),
+	NETWORK_NEAR: str({ choices: ["mainnet", "testnet"] }),
+	BRIDGE_TOKEN_FACTORY_ACCOUNT_ID: str(),
+});
```
olga24912, line 27: We should add INFURA_API_KEY here. 

```
/home/olga/Aurora/MainnetMigration/AuroraEthConnectorMainnetMigration/omni-bridge-migration/migration/checkERC20EthToNearTransfersFinalization/node_modules/.pnpm/envalid@8.0.0/node_modules/envalid/src/middleware.ts:49
          throw new ReferenceError(
                ^


ReferenceError: [envalid] Env var INFURA_API_KEY was accessed but not validated. This var is set in the environment; please add an envalid validator for it.
```

olga24912 (changes requested):

migration/checkERC20EthToNearTransfersFinalization/index.ts
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/290/commits/44613661451caf79f9f7fb138d639fb57734b6f7

olga24912 (approved):

PR #297, opened 2025-03-13
fix: finalise legacy transfers

karim-en:
Allow the e-near proxy contract to finalize the legacy transfers

### DIFF
--- a/evm/src/eNear/contracts/ENearProxy.sol
+++ b/evm/src/eNear/contracts/ENearProxy.sol
@@ -12,6 +12,7 @@ contract ENearProxy is UUPSUpgradeable, AccessControlUpgradeable, ICustomMinter,
     IENear public eNear;
 
     bytes32 public constant MINTER_ROLE = keccak256("MINTER_ROLE");
+    bytes32 public constant PAUSE_ROLE = keccak256("PAUSE_ROLE");
     bytes public nearConnector;
     uint256 public currentReceiptId;
     INearProver public prover;
@@ -31,6 +32,7 @@ contract ENearProxy is UUPSUpgradeable, AccessControlUpgradeable, ICustomMinter,
         currentReceiptId = _currentReceiptId;
         prover = INearProver(_prover);
         _grantRole(DEFAULT_ADMIN_ROLE, _adminAddress);
+        _grantRole(PAUSE_ROLE, _msgSender());
     }
 
     function mint(address token, address to, uint128 amount) public onlyRole(MINTER_ROLE) {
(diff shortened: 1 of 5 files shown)


evm/src/eNear/contracts/ENearProxy.sol
```diff
@@ -53,6 +58,22 @@ contract ENearProxy is UUPSUpgradeable, AccessControlUpgradeable, ICustomMinter
...
+    ) external whenNotPaused(PAUSED_LEGACY_FIN_TRANSFER) {
+        require(
+            prover.proveOutcome(proofData, proofBlockHeight),
+            "Proof should be valid"
+        );
+
+        eNear.finaliseNearToEthTransfer(proofData, proofBlockHeight);
+    }
+
+    function pause(uint flags) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
karim-en, line 73: Let's also add `pauseAll` with pause manager role

kiseln: 61aaa7c67d6d76789e1f8ed216776b4e34f80681

kiseln (approved):

evm/src/eNear/contracts/ENearProxy.sol
karim-en: We should unify the naming of the role between our EVM contracts.
https://github.com/Near-One/omni-bridge/blob/main/evm%2Fsrc%2Fomni-bridge%2Fcontracts%2FOmniBridge.sol#L37

kiseln: Done

kiseln (approved):

evm/src/eNear/contracts/ENearProxy.sol
```diff
@@ -53,6 +60,26 @@ contract ENearProxy is UUPSUpgradeable, AccessControlUpgradeable, ICustomMinter
...
+    function finaliseNearToEthTransfer(
+        bytes memory proofData,
+        uint64 proofBlockHeight
+    ) external whenNotPaused(PAUSED_LEGACY_FIN_TRANSFER) {
+        require(
+            prover.proveOutcome(proofData, proofBlockHeight),
+            "Proof should be valid"
+        );
+
+        eNear.finaliseNearToEthTransfer(proofData, proofBlockHeight);
```
olga24912, line 72: Why do we need to add a separate method to this contract? Why can't we finalize transactions directly in eNear?

olga24912 (approved):

PR #300, opened 2025-03-18
feat: added script to verify contract hash

frolvanya:
closes https://github.com/Near-One/omni-bridge-private/issues/37


kiseln:
https://github.com/Near-One/omni-bridge-private/issues/35

PR #308, opened 2025-04-03
chore: lowercase and title-case serialization

frolvanya:
This is a request from Igor Kazakov and overall I think it would be very neat for UX. Right now we have to specify chain_kind in "Title case" in api, sdk, rpc

closes https://github.com/Near-One/bridge-indexer-rs/issues/142

### DIFF
--- a/near/omni-types/src/lib.rs
+++ b/near/omni-types/src/lib.rs
@@ -142,12 +142,18 @@ impl Serialize for H160 {
     IntoPrimitive,
 )]
 #[repr(u8)]
+#[serde(rename_all = "lowercase")]
 pub enum ChainKind {
     #[default]
+    #[serde(alias = "Eth")]
     Eth,
+    #[serde(alias = "Near")]
     Near,
+    #[serde(alias = "Sol")]
     Sol,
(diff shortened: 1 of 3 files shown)


kiseln (approved):

near/omni-types/src/lib.rs
```diff
@@ -144,10 +144,15 @@ impl Serialize for H160 {
 #[repr(u8)]
 pub enum ChainKind {
     #[default]
+    #[serde(alias = "eth")]
```
olga24912, line 147: We use this type in many places, and I thought it was also used in contracts. Won’t we break backward compatibility if we change the serialization method now?

For example, will valid contract calls still be generated?

olga24912 (changes requested):

near/omni-types/src/lib.rs
frolvanya: it's not a rename, it's alias, given that both `Eth` and `eth` options will work, so backward compatibility remains

olga24912 (approved):

karim-en (approved):

PR #313, opened 2025-04-11
fix: fast transfers state migration

karim-en:


near/omni-types/src/lib.rs
```diff
@@ -438,6 +438,19 @@ pub struct TransferId {
     pub origin_nonce: Nonce,
 }
 
+#[near(serializers=[borsh, json])]
+#[derive(Debug, Clone)]
+pub struct TransferMessageV0 {
```
frolvanya, line 443: Do we really need a V0 structure for this, if in original structure `origin_transfer_id` is optional?

frolvanya (approved):

near/omni-bridge/src/migrate.rs
```diff
@@ -0,0 +1,50 @@
...
+    #[init(ignore_state)]
+    pub fn migrate() -> Self {
+        let old_state: OldState = env::state_read().expect("failed");
+
+        Self {
+            prover_account: old_state.prover_account,
+            factories: old_state.factories,
+            pending_transfers: old_state.pending_transfers,
+            finalised_transfers: old_state.finalised_transfers,
+            fast_transfers: LookupMap::new(StorageKey::FastTransfers),
```
karim-en, line 37: The order of the storage keys cann't be changed

karim-en: fixed in d63ea59c0e9c1072d22cff42263ec4aa742f57ab

karim-en:
The integration test should be improved to cover the whole storage state

PR #315, opened 2025-04-17
feat: added aurora chain kind and couple events for indexing

frolvanya:
I'm still not sure if we should add aurora `ChainKind`/`OmniAddress` into omni-types or it's better to hard code it in bridge-indexer somehow


frolvanya:
We decided that it would be better to create a separate indexer called `evm-on-near-indexer` that wouldn't rely on `ChainKind::Aurora`

PR #317, opened 2025-04-21
feat: migration to lake-0.8.0

frolvanya:
Right now relayer is stuck due to the new actions related to global contracts


frolvanya:
closing in favour of #318

PR #327, opened 2025-05-11
feat: update decimals over `set_token_metadata`

karim-en:
Some of the migrated Rainbow tokens don't have the metadata, for example, this one https://github.com/aurora-is-near/bridge-assets/pull/326.
This PR allows setting decimal by DAO over the `set_token_metadata`

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -1012,23 +1012,33 @@ impl Contract {
     #[access_control_any(roles(Role::DAO, Role::MetadataManager))]
     pub fn set_token_metadata(
         &mut self,
-        token: AccountId,
+        address: OmniAddress,
         name: Option<String>,
         symbol: Option<String>,
         icon: Option<String>,
         reference: Option<String>,
         reference_hash: Option<Base64VecU8>,
-        decimals: Option<u8>,
     ) -> Promise {
-        // Only DAO can set decimals
-        require!(
-            decimals.is_none()
-                || self.acl_has_role(Role::DAO.into(), env::predecessor_account_id())
-        );
+        let token = self.get_token_id(&address);
+        require!(self.deployed_tokens.contains(&token));
+
+        let decimals = self
+            .token_decimals
+            .get(&address)
+            .sdk_expect("ERR_TOKEN_DECIMALS_NOT_FOUND")
+            .decimals;
+        require!(decimals > 0, "ERR_INVALID_DECIMALS");
 
         ext_token::ext(token)
             .with_static_gas(SET_METADATA_GAS)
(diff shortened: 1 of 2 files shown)


near/omni-bridge/src/lib.rs
```diff
@@ -1018,10 +1018,17 @@ impl Contract {
         icon: Option<String>,
         reference: Option<String>,
         reference_hash: Option<Base64VecU8>,
+        decimals: Option<u8>,
     ) -> Promise {
+        // Only DAO can set decimals
+        require!(
+            decimals.is_none()
+                || self.acl_has_role(Role::DAO.into(), env::predecessor_account_id())
```
frolvanya, line 1026: Looks a bit strange to me. Isn't is better to create a separate method called `set_decimals` with only DAO role to avoid confusion?

frolvanya (approved):

near/omni-bridge/src/lib.rs
karim-en: It could be, but creating two proposals to update the metadata isn't convenient.

olga24912 (approved):
Do we currently have any tokens without decimals in OmniBridge? Are there any issues with conversion when transferring them from one blockchain to another?

If there are many such tokens, it will be hard for us to manually fix all the decimals through the DAO.

karim-en:
> Do we currently have any tokens without decimals in OmniBridge? Are there any issues with conversion when transferring them from one blockchain to another?
> 
> If there are many such tokens, it will be hard for us to manually fix all the decimals through the DAO.

@olga24912 29 tokens with 0 decimals

olga24912 (approved):

near/omni-bridge/src/lib.rs
```diff
@@ -1012,16 +1012,33 @@ impl Contract {
...
     ) -> Promise {
+        let token = self.get_token_id(&address);
+        require!(self.deployed_tokens.contains(&token));
+
+        let decimals = self
+            .token_decimals
+            .get(&address)
+            .sdk_expect("ERR_TOKEN_DECIMALS_NOT_FOUND")
+            .decimals;
+        require!(decimals > 0, "ERR_INVALID_DECIMALS");
```
frolvanya, line 1030: What if decimals in internal map is 0? Is it possible? How to update decimals in this case?

frolvanya (approved):

near/omni-bridge/src/lib.rs
karim-en: It's an abnormal scenario, I hope we didn't make a mistake like this during the migration.
But it is a good question, because it looks like it is ok to have token zero decimals, and we should allow to do that.

karim-en: 2012699d7f454a085dcfee728812f6b31bd65714

PR #337, opened 2025-05-25
chore: bumped alloy crate

frolvanya:


near/omni-types/src/evm/events.rs
```diff
@@ -58,11 +56,7 @@ where
     <V as TryFromLog<Log<T>>>::Error: std::fmt::Display,
 {
     let rlp_decoded = Log::decode(&mut log_rlp.as_slice()).map_err(stringify)?;
-    V::try_from_log(
-        chain_kind,
-        T::decode_log(&rlp_decoded, true).map_err(stringify)?,
-    )
-    .map_err(stringify)
+    V::try_from_log(chain_kind, T::decode_log(&rlp_decoded).map_err(stringify)?).map_err(stringify)
```
karim-en, line 59: Looks like the decode now is called without validation https://docs.rs/alloy-sol-types/latest/src/alloy_sol_types/types/event/mod.rs.html#221

frolvanya: Will be probably fixed in https://github.com/alloy-rs/core/issues/955

karim-en (approved):

near/omni-types/src/evm/events.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/337/commits/23b339402dc1e993c7abebdd374ba6fc20b1c013

PR #338, opened 2025-05-26
feat: sign btc transaction

frolvanya:


olga24912 (dismissed):
Am I right in thinking that I’ve already reviewed all this code? What happened with the branches — why didn’t it end up in master?

olga24912 (approved):

olga24912 (approved):

PR #342, opened 2025-06-11
feat: support for fin transfers"

frolvanya:


omni-relayer/src/config.rs
```diff
@@ -6,9 +6,17 @@ use near_primitives::types::AccountId;
 use omni_types::{ChainKind, OmniAddress};
 use serde::Deserialize;
 
-pub fn get_private_key(chain_kind: ChainKind) -> String {
+pub enum NearSignerType {
+    Omni,
+    Fast,
+}
+
+pub fn get_private_key(chain_kind: ChainKind, near_signer_type: Option<NearSignerType>) -> String {
```
frolvanya, line 14: Right now it's a bit strange to pass `None` for `near_signer_type` if we're getting private key not for NEAR. Let me know if it's better to do something like:

```rust
enum GetPrivateKeyParams {
    Near(NearSignerType),
    Other(ChainKind)
}
```

omni-relayer/src/main.rs
```diff
@@ -54,16 +54,39 @@ async fn main() -> Result<()> {
...
-    let near_signer = startup::near::get_signer(config.near.credentials_path.as_ref())?;
 
-    let connector = Arc::new(startup::build_omni_connector(&config, &near_signer)?);
+    let near_omni_signer = startup::near::get_signer(
+        config.near.omni_credentials_path.as_ref(),
+        config::NearSignerType::Omni,
+    )?;
 
-    let near_nonce = Arc::new(utils::nonce::NonceManager::new(
+    let near_fast_signer = if config.is_fast_relayer_enabled() {
```
frolvanya, line 63: Should I combine them like I did with EVM?

omni-relayer/src/main.rs
```diff
@@ -54,16 +54,39 @@ async fn main() -> Result<()> {
...
+        Some(startup::near::get_signer(
+            config.near.fast_credentials_path.as_ref(),
+            config::NearSignerType::Fast,
+        )?)
+    } else {
+        None
+    };
+    let (connector, near_fast_bridge_client) =
```
frolvanya, line 71: It's a bit strange to have a separate bridge client for fast relayer, let me know if you have better ideas to resolve this. Mostly I need this connector just to call `fast_fin_transfer` method

olga24912 (commented):
Why do you want to merge this into a branch related to BTC? This functionality doesn’t seem related at all.

frolvanya:
> Why do you want to merge this into a branch related to BTC? This functionality doesn’t seem related at all.

I just didn't want to resolve conflicts afterwards, since all PRs at the end will be merged in main, IMO it's better to streamline the process with stacking PRs on top of each other

omni-relayer/src/main.rs
frolvanya: After discussion with @karim-en we decided that it's much better just to have 2 omni-connector structures (1 for regular connector and the other one will contain different near-bridge-client for a fast relayer)

omni-relayer/src/workers/evm.rs
```diff
@@ -159,10 +164,65 @@ pub async fn process_init_transfer_event(
...
+            .near_get_fast_transfer_status(fast_transfer_args.id())
+            .await
+        else {
+            warn!("Failed to get fast transfer status for transfer: {transfer_id:?}");
+            return Ok(EventAction::Retry);
+        };
+
+        if let Some(status) = fast_transfer_status {
+            recipient = OmniAddress::Near(status.relayer.clone());
+            fee_recipient = status.relayer;
```
karim-en, line 210: Does it still work if the fast relayer isn't enabled? I mean, if we run just the regular relayer?

frolvanya: Fixed in 6498708c68d992afbd264475d70942115a2ec0cc

karim-en (approved):

omni-relayer/src/config.rs
frolvanya: Because for other chains we don't need to specify near_signer_type. Only for NEAR we need to do it

omni-relayer/src/config.rs
```diff
@@ -146,6 +161,8 @@ pub struct Evm {
     pub light_client: Option<AccountId>,
     pub block_processing_batch_size: u64,
     pub expected_finalization_time: i64,
+    #[serde(default)]
```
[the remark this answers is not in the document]
karim-en: Why u64::max_value?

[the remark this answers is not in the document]
frolvanya: To prevent issue where we set `is_fast_relayer_enabled` to true, but forgetting to specify `safe_confirmation` (this would default to 0 in previous implementation, meaning that this would be too dangerous)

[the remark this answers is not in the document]
frolvanya: Or do you mean why I chose deprecated `u64::max_value` over `u64::MAX`? Because `default` field asks for a function, so I either need to create own function that returns `u64::MAX` or use this deprecated implementation

kiseln (approved):

PR #347, opened 2025-06-17
Allow any relayer to finalize fast transfer

kiseln:
## Explanation
Previously, only the relayer that executed a fast transfer could finalize it after confirmation. This restriction made sense since that relayer earned the fees and had the primary incentive to complete the process. However, some 3rd parties may want to run a multi-relayer setup where one relayer performs fast transfers while another handles finalization. To support this scenario we remove this restriction and allow any account to finalize the transfer (the fees are still being sent to the fast relayer).

## Implementation
- Remove the restriction on fast transfer finalization
- In case fast transfer exists, change fee recipient to the fast relayer
- Original caller still pays for storage


karim-en (changes requested):
What about the `process_fin_transfer_to_other_chain`, can we allow any relayer to perform that?

kiseln:
> What about the `process_fin_transfer_to_other_chain`, can we allow any relayer to perform that?

Added. On tests, I did a little revamp of those in #349 so would rather add them after it is merged

frolvanya (approved):
Maybe it would be better to give an option to fast relayer to explicitly say that he wants to finalize all by himself? On the other hand, I don't think anyone will want to do this and regular case would be to let `omni-relayer` to do the rest

olga24912 (approved):

karim-en (approved):

kiseln:
Added tests. They are getting a bit messy so I intend to refactor at some point

karim-en (approved):

PR #349, opened 2025-06-18
Safety check for fast transfer amount

kiseln:
## Summary
In this PR we add an additional field called `origin_amount` to the Fast Transfer message. It serves no behavioral purpose but acts as an additional safety check to make sure the amount submitted for the fast transfer is correct. This is important because, in case of a mistake, the relayer will not be able to recover the funds when transfer reaches finality.

## Issues

For consistency we can also change the `fee` property of the message to contain normalized amount. I didn't do it to keep the change less invasive but I believe it's worth considering

### DIFF
--- a/near/omni-tests/src/fast_transfer.rs
+++ b/near/omni-tests/src/fast_transfer.rs
@@ -43,7 +43,7 @@ mod tests {
 
         #[allow(clippy::too_many_lines)]
         async fn new(is_bridged_token: bool) -> anyhow::Result<Self> {
-            let sender_balance_token = 1_000_000;
+            let sender_balance_token = 1_000_000_000_000;
             let worker = near_workspaces::sandbox().await?;
 
             let prover_contract = worker.dev_deploy(&mock_prover_wasm()).await?;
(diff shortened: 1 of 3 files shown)


near/omni-tests/src/fast_transfer.rs
```diff
@@ -800,17 +843,19 @@ mod tests {
...
+            let (_, fast_transfer_msg) = get_transfer_to_other_chain_msg(&env, transfer_amount, fee, decimal_diff);
+ 
             let relayer_balance_before =
                 get_balance(&env.token_contract, env.relayer_account.id()).await?;
             let contract_balance_before =
                 get_balance(&env.token_contract, env.bridge_contract.id()).await?;
 
             let result = do_fast_transfer(&env, transfer_amount, fast_transfer_msg.clone()).await?;
 
+            println!("result: {:?}", result);
```
karim-en, line 858: Let's remove this print

karim-en (changes requested):

karim-en:
Please fic the CI fmt/clippy

near/omni-tests/src/fast_transfer.rs
kiseln: done

PR #351, opened 2025-06-23
Pass fast transfer fee as it is on origin chain

kiseln:
It will be easier for contract users to submit amounts in `FastTransferMsg` as they are on origin chain. We will convert it to Near decimals in the contract while also validating that the amount passed is correct

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -608,10 +608,7 @@ impl Contract {
             token: init_transfer.token,
             amount: Self::denormalize_amount(init_transfer.amount.0, decimals).into(),
             recipient: init_transfer.recipient,
-            fee: Fee {
-                fee: Self::denormalize_amount(init_transfer.fee.fee.0, decimals).into(),
-                native_fee: init_transfer.fee.native_fee,
-            },
+            fee: Self::denormalize_fee(init_transfer.fee, decimals),
             sender: init_transfer.sender,
             msg: init_transfer.msg,
             destination_nonce,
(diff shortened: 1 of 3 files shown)


near/omni-bridge/src/lib.rs
```diff
@@ -1848,4 +1846,12 @@ impl Contract {
         let diff_decimals: u32 = (decimals.origin_decimals - decimals.decimals).into();
         amount / (10_u128.pow(diff_decimals))
     }
+
+    // Native tokens always have the same decimals on Near as on origin chain
+    fn denormalize_fee(fee: Fee, decimals: Decimals) -> Fee {
```
frolvanya, line 1851: Is there a point of creating a separate method for this? If there's a need for it, then maybe it's better add it as a method for a `Fee` structure in `omni-types`?

near/omni-bridge/src/lib.rs
```diff
@@ -650,18 +647,19 @@ impl Contract {
             .get(&origin_token)
             .sdk_expect("ERR_TOKEN_DECIMALS_NOT_FOUND");
 
-        let total_amount =
-            Self::denormalize_amount(fast_fin_transfer_msg.origin_amount.0, decimals);
+        let denormalized_amount =
```
frolvanya, line 650: So now we're passing amount and fee without denormalization. Am I understanding it correctly? So we need to fix it in sdk-rs/sdk-js

kiseln: This is correct

olga24912 (approved):

karim-en (approved):

PR #353, opened 2025-06-26
feat: remember historical fees

frolvanya:
image

### DIFF
--- a/omni-relayer/Cargo.lock
+++ b/omni-relayer/Cargo.lock
@@ -2107,7 +2107,7 @@ dependencies = [
 [[package]]
 name = "bridge-connector-common"
 version = "0.2.1"
-source = "git+https://github.com/Near-One/bridge-sdk-rs?rev=7d381da690f28bd479ccc78695aceaffa6467ba1#7d381da690f28bd479ccc78695aceaffa6467ba1"
+source = "git+https://github.com/Near-One/bridge-sdk-rs?rev=d0ee564e57197eac92fae45b7a021d107e662cf3#d0ee564e57197eac92fae45b7a021d107e662cf3"
 dependencies = [
  "eth-proof",
  "ethers",
@@ -2122,7 +2122,7 @@ dependencies = [
 [[package]]
 name = "bridge-indexer-types"
 version = "0.1.0"
-source = "git+ssh://git@github.com/near-one/bridge-indexer-rs.git?rev=f20669ebf30e191db5dcf1656724917b69b9a0b0#f20669ebf30e191db5dcf1656724917b69b9a0b0"
+source = "git+ssh://git@github.com/near-one/bridge-indexer-rs.git?rev=ed969be97bc9cdb3bf55d96cee213b342911363b#ed969be97bc9cdb3bf55d96cee213b342911363b"
 dependencies = [
  "borsh 1.5.7",
  "bson",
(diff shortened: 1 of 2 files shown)


omni-relayer/src/utils/redis.rs
```diff
@@ -17,6 +21,41 @@ pub const SLEEP_TIME_AFTER_EVENTS_PROCESS_SECS: u64 = 10;
 const QUERY_RETRY_ATTEMPTS: u64 = 10;
 const QUERY_RETRY_SLEEP_SECS: u64 = 1;
 
+pub async fn get_fee(
+    redis_connection: &mut MultiplexedConnection,
+    origin_nonce: u64,
+) -> Option<TransferFeeResponse> {
+    for _ in 0..QUERY_RETRY_ATTEMPTS {
+        match redis_connection
+            .hget::<&str, u64, Option<String>>(FEE_MAPPING, origin_nonce)
```
frolvanya, line 30: I understand that there's a possibility of `origin_nonce` collision and it's better to store it as a `TransferId`, but:
1) `TransferId` needs to be converted to string using `serde_json`, so it complicates code a little bit
2) I'll rewrite it to RabbitMQ soon, so I don't see the reason to perfect current solution. This can be considered as a temporary solution to avoid issues with long signs (e.g due to issues with MPC)

omni-relayer/src/workers/mod.rs
```diff
@@ -84,6 +84,12 @@ pub enum Transfer {
     },
 }
 
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
+pub struct SignTransferEvent {
+    pub event: OmniBridgeEvent,
+    pub signer_id: AccountId,
```
frolvanya, line 90: @karim-en I remember that we were planning to check `predecessor_id`, but for sign transfer event, predecessor is omni-bridge contract, so I decided to go with `signer_id`

image

I think that this is safe, because if transaction was initiated by our relayer, then fee is sufficient, but let me know if this check could result in malicious free transfer finalization somehow

omni-relayer/src/utils/redis.rs
frolvanya: Nvm, decided that it's better to fix it:
https://github.com/Near-One/omni-bridge/pull/353/commits/ada4820e2b95229512cfa44060b33cddaba52132

karim-en (approved):
LGTM!

omni-relayer/src/workers/mod.rs
karim-en: not relevant

frolvanya:
We forgot that relayer treats transfers from aurora differently and we sign transfers without checking fee. So that means that previous implementation in which we dropped fee check after sign is received (if sign is from our relayer) meant that any transfer from aurora would be finalized without any fee check

omni-relayer/src/utils/bridge_api.rs
```diff
@@ -29,28 +31,72 @@ pub async fn is_fee_sufficient(
...
         .await?
         .json::<TransferFeeResponse>()
-        .await?;
+        .await
+        .map_err(Into::into)
+}
 
-    let native_fee = response.native_token_fee.unwrap_or_default().0;
-    let transferred_fee = response.transferred_token_fee.unwrap_or_default().0;
+pub fn is_fee_sufficient(
```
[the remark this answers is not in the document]
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/353/commits/722acf28f0fe8a98b0327ff2ea8c328916cbff83

PR #364, opened 2025-07-04
feat: removed `add_provers` method

frolvanya:
Removed since it can be done using batches


karim-en:
Please bump the prover version

karim-en (approved):

PR #369, opened 2025-07-15
feat: added flag for `verify_withdraw` call

frolvanya:


omni-relayer/example-devnet-config.toml
```diff
@@ -92,6 +92,8 @@ finalize_transfer_sol_discriminator = [104, 27, 121, 69, 3, 70, 217, 66]
 rpc_http_url = "https://bitcoin-testnet-rpc.publicnode.com"
 # If this is set to true, then omni relayer will call `sign_btc_transaction`
 signing_enabled = false
+# If this is set to true, then omni relayer will call `verify_withdraw`
+verifying_withdraw_enabled = false
```
kiseln, line 96: It is not very clear from this value that this is UTXO related. I would try something descriptive like `near_to_utxo_transfers_enabled`

PR #370, opened 2025-07-25
feat: implement refund

karim-en:
This PR adds refund logic in case the `ft_on_transfer` fails

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -65,7 +65,7 @@ const BURN_TOKEN_GAS: Gas = Gas::from_tgas(3);
 const MINT_TOKEN_GAS: Gas = Gas::from_tgas(5);
 const SET_METADATA_GAS: Gas = Gas::from_tgas(10);
 const RESOLVE_FAST_TRANSFER_GAS: Gas = Gas::from_tgas(6);
-const FAST_TRANSFER_CALLBACK_GAS: Gas = Gas::from_tgas(5);
+const FAST_TRANSFER_CALLBACK_GAS: Gas = Gas::from_tgas(10);
 const NO_DEPOSIT: NearToken = NearToken::from_near(0);
 const ONE_YOCTO: NearToken = NearToken::from_yoctonear(1);
 const SEND_TOKENS_CALLBACK_GAS: Gas = Gas::from_tgas(15);
@@ -726,10 +726,12 @@ impl Contract {
             .to_log_string(),
         );
 
+        let amount = U128(fast_transfer.amount.0 - fast_transfer.fee.fee.0);
+
         self.send_tokens(
             fast_transfer.token_id.clone(),
             recipient,
-            U128(fast_transfer.amount.0 - fast_transfer.fee.fee.0),
+            amount,
             &fast_transfer.msg,
         )
         .then(
@@ -738,7 +740,7 @@ impl Contract {
                 .resolve_fast_transfer(
                     fast_transfer.token_id.clone(),
                     &fast_transfer.id(),
-                    fast_transfer.amount,
+                    amount,
                     !fast_transfer.msg.is_empty(),
                 ),
         )
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -1353,6 +1353,7 @@ impl Contract {
         #[serializer(borsh)] transfer_message: TransferMessage,
         #[serializer(borsh)] fee_recipient: AccountId,
         #[serializer(borsh)] is_ft_transfer_call: bool,
+        #[serializer(borsh)] storage_owner: AccountId,
     ) {
         let is_refund = if is_ft_transfer_call {
             match env::promise_result(0) {
@@ -1374,9 +1375,7 @@ impl Contract {
         };
 
         if is_refund {
-            // TODO: handle storage update
-            self.finalised_transfers
-                .remove(&transfer_message.get_transfer_id());
+            self.remove_fin_transfer(&transfer_message.get_transfer_id(), &storage_owner);
         } else {
             // Send fee to the fee recipient
             let token = self.get_token_id(&transfer_message.token);
@@ -1500,7 +1499,7 @@ impl Contract {
         }
 
         self.update_storage_balance(
-            predecessor_account_id,
+            predecessor_account_id.clone(),
             required_balance,
             env::attached_deposit(),
         );
@@ -1511,9 +1510,10 @@ impl Contract {
                 Self::ext(env::current_account_id())
                     .with_static_gas(SEND_TOKENS_CALLBACK_GAS)
                     .fin_transfer_send_tokens_callback(
-                        transfer_message.clone(),
-                        fee_recipient.clone(),
+                        transfer_message,
+                        fee_recipient,
                         !msg.is_empty(),
+                        predecessor_account_id,
                     ),
             )
     }
@@ -1759,6 +1759,19 @@ impl Contract {
             storage.available = storage.available.saturating_add(refund);
             self.accounts_balances
                 .insert(&fast_transfer.storage_owner, &storage);
+        }
+    }
+
+    fn remove_fin_transfer(&mut self, transfer_id: &TransferId, storage_owner: &AccountId) {
+        let storage_usage = env::storage_usage();
+        self.finalised_transfers.remove(transfer_id);
+
+        let refund =
+            env::storage_byte_cost().saturating_mul((storage_usage - env::storage_usage()).into());
+
+        if let Some(mut storage) = self.accounts_balances.get(storage_owner) {
+            storage.available = storage.available.saturating_add(refund);
+            self.accounts_balances.insert(&storage_owner, &storage);
         }
     }
 
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -168,13 +168,6 @@ pub trait TokenDeployer {
     fn deploy_token(&self, account_id: AccountId, metadata: BasicMetadata) -> Promise;
 }
 
-pub struct RefundData {
-    pub fast_transfer: Option<FastTransfer>,
-    pub transfer_message: TransferMessage,
-    pub message_owner: AccountId,
-    pub is_refund_required: bool,
-}
-
 #[near(contract_state)]
 #[derive(Pausable, Upgradable, PanicOnDefault)]
 #[access_control(role_type(Role))]
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -1413,9 +1413,9 @@ impl Contract {
                     .with_static_gas(MINT_TOKEN_GAS)
                     .mint(fee_recipient.clone(), transfer_message.fee.native_fee, None);
             }
-        }
 
-        env::log_str(&OmniBridgeEvent::FinTransferEvent { transfer_message }.to_log_string());
+            env::log_str(&OmniBridgeEvent::FinTransferEvent { transfer_message }.to_log_string());
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


near/omni-bridge/src/lib.rs
```diff
@@ -1345,6 +1346,67 @@ impl Contract {
...
+            if transfer_message.fee.native_fee.0 > 0 {
+                let native_token_id = self.get_native_token_id(transfer_message.get_origin_chain());
+
+                ext_token::ext(native_token_id)
+                    .with_static_gas(MINT_TOKEN_GAS)
+                    .mint(fee_recipient.clone(), transfer_message.fee.native_fee, None);
+            }
+        }
+
+        env::log_str(&OmniBridgeEvent::FinTransferEvent { transfer_message }.to_log_string());
```
karim-en, line 1408: Mark it as a refund?

frolvanya: Well, it doesn't really matter for relayer since it's only for transfers to near and not other chains and relayer simply ignores them. But it'd be nice for checking if transfer was refunded in indexers (and optionally to send some alerts for these cases)

karim-en: Yeh, we have two options here:
- Add refunded field in the FinTransferEvent
- Don't emit an event if transfer refunded

frolvanya: We also can emit a different log (e.g `RefundedFinTransfer`)

Pros:
- We don't break existing `FinTransferEvent`-s stored in db
- We still can add some sort of alerts for this case
- We don't break API

Cons:
- Might be a bit confusing to have two identical structures

But completely disabling event means that we won't be able to track such events later. AFAIK refunds are still considered as a "rare" event, so I'd like to configure alerts for them

karim-en:
Added tests in 5906cbc3a2848b6e9657b0ae7e368f0a40b346c9

near/omni-tests/src/fin_transfer.rs
```diff
@@ -201,15 +239,20 @@ mod tests {
     }
 
     #[allow(clippy::too_many_lines)]
-    async fn test_fin_transfer(
-        storage_deposit_accounts: Vec<(AccountId, bool)>,
+    async fn internal_test_fin_transfer(
```
karim-en, line 242: This test doesn't include the mint flow

karim-en: Fixed in 1d60ef22aff0687f74e9c8fa0e83d7897bd4333c

near/omni-bridge/src/lib.rs
```diff
@@ -1345,6 +1346,67 @@ impl Contract {
...
+                // Unexpected case: don't refund
+                PromiseResult::Failed => false,
+            }
+        } else {
+            // Not ft_transfer_call: don't refund
+            false
+        };
+
+        if is_refund {
+            self.remove_fin_transfer(&transfer_message.get_transfer_id(), storage_owner);
```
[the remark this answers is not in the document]
karim-en: Fixed in 1d60ef22aff0687f74e9c8fa0e83d7897bd4333c

near/omni-bridge/src/lib.rs
```diff
@@ -1345,6 +1347,76 @@ impl Contract {
...
+
+            if transfer_message.fee.native_fee.0 > 0 {
+                let native_token_id = self.get_native_token_id(transfer_message.get_origin_chain());
+
+                ext_token::ext(native_token_id)
+                    .with_static_gas(MINT_TOKEN_GAS)
+                    .mint(fee_recipient.clone(), transfer_message.fee.native_fee, None);
+            }
+
+            env::log_str(&OmniBridgeEvent::FinTransferEvent { transfer_message }.to_log_string());
```
frolvanya, line 1417: I still think that adding a new type or optional field would be better (in order to keep current events without reindexing, but it can be also done simply by adding `refunded: false` field to all current events)

near/omni-bridge/src/lib.rs
```diff
@@ -1345,6 +1347,76 @@ impl Contract {
...
+            if is_deployed_token {
+                let amount_to_burn = U128(transfer_message.amount.0 - transfer_message.fee.fee.0);
+
+                ext_token::ext(token)
+                    .with_static_gas(BURN_TOKEN_GAS)
+                    .burn(amount_to_burn);
+            }
+
+            self.remove_fin_transfer(&transfer_message.get_transfer_id(), storage_owner);
+        } else {
```
frolvanya, line 1391: Nitpick: This can be simplified by adding return statement to reduce indentation level

```
if ... {
   ...
   return;
}

...
```

instead of

```
if ... {
   ...
} else {
   ...
}
```

frolvanya (approved):
Also, just to double check. In case of fast transfer, who will receive the refund? And what should we do in this case? Retry finalization to return tokens to fast relayer?

near/omni-bridge/src/lib.rs
karim-en: f7fb3a639e5e7d829e61bb3c574b33e8370df9a3

near/omni-bridge/src/lib.rs
```diff
@@ -684,29 +664,30 @@ impl Contract {
...
+                    ),
             )
+            .into()
         } else {
             self.fast_fin_transfer_to_other_chain(
                 &fast_transfer,
                 storage_payer,
                 fast_fin_transfer_msg.relayer,
             );
+            self.burn_tokens_if_needed(token_id, amount);
```
frolvanya, line 690: I don't understand why do we need to burn tokens after fast transfer. Shouldn't it be done after regular transfer? Also, what happens in normal fast transfer case? Won't we burn tokens twice? First burn will happen in fast transfer and then in a regular finalisation

frolvanya (approved):

near/omni-bridge/src/lib.rs
```diff
@@ -471,7 +449,7 @@ impl Contract {
         token_id: AccountId,
         amount: U128,
         init_transfer_msg: InitTransferMsg,
-    ) -> U128 {
+    ) -> PromiseOrValue<U128> {
```
r-near, line 452: Why change the return type if we're always returning the `Value` variant?

near/omni-bridge/src/lib.rs
```diff
@@ -1452,23 +1494,26 @@ impl Contract {
...
-            predecessor_account_id,
+            predecessor_account_id.clone(),
             required_balance,
             env::attached_deposit(),
         );
 
-        env::log_str(&OmniBridgeEvent::FinTransferEvent { transfer_message }.to_log_string());
-
-        promise
+        let amount_to_transfer = U128(transfer_message.amount.0 - transfer_message.fee.fee.0);
```
r-near, line 1505: Probably unlikely, but we should check for underflow here, for potential precision loss issues

near/omni-bridge/src/lib.rs
```diff
@@ -1345,9 +1327,87 @@ impl Contract {
...
+                    }
+                }
+                // Unexpected case: don't refund
+                PromiseResult::Failed => false,
+            }
+        } else {
+            // Not ft_transfer_call: don't refund
+            false
+        }
+    }
```
r-near, line 1401: Based on [NEP-141](https://nomicon.io/Standards/Tokens/FungibleToken/Core), `ft_on_transfer` is supposed to return unused tokens, where 0 means success and >0 means refund needed.

Looking at this code:
```rust
amount.0 == 0  // Currently returns true when 0 unused tokens
``` 

Should this be amount.0 > 0 instead? It seems like the current logic would treat successful transfers (0 unused) as failures and vice versa?  

Also wondering if we should default to refund on parsing errors for safety

karim-en: - it is a little bit confusing, but the `ft_trnasfer_call` returns the used amount not the unused.
- for safety we shouldn't refund by default, the default NEP141 implementation has some undetermined behaviour, where the failed call can lead to two different results. 
1- the recipient receives the tokens
2- the recipient doesn't receive the tokens and the tokens refunded back to the sender.

The problem is that there is no way to distinguish between the errors types in the callback

near/omni-bridge/src/lib.rs
karim-en: We have two levels of protection for this:
- We don't accept transfers with `fee >= amount`
- The Near contract is compiled with the flag `overflow-checks = true`

So we have this rule:
- Use checked math operations when the overflow is possible, eg, validating user input.
- If the overflow isn't expected, then use regular operations, but the flag `overflow-checks = true` should be enabled to protect us from mistakes.

near/omni-bridge/src/lib.rs
```diff
@@ -752,8 +733,13 @@ impl Contract {
         )
         .then(
             Self::ext(env::current_account_id())
-                .with_static_gas(RESOLVE_TRANSFER_GAS)
-                .resolve_transfer(),
+                .with_static_gas(RESOLVE_FAST_TRANSFER_GAS)
+                .resolve_fast_transfer(
```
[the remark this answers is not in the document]
karim-en: it works for any, but the refund flow is enabled only when the msg isn't empty.

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
karim-en: yep

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
kiseln: For simplicity we can pass the whole `fast_transfer` struct, I don't think the gas difference will be noticeable

near/omni-bridge/src/lib.rs
```diff
@@ -1307,23 +1293,19 @@ impl Contract {
...
+        token_id: AccountId,
+        fast_transfer_id: &FastTransferId,
+        amount: U128,
+        is_ft_transfer_call: bool,
+    ) -> U128 {
+        if Self::is_refund_required(is_ft_transfer_call) {
+            self.remove_fast_transfer(fast_transfer_id);
+            amount
+        } else {
+            self.burn_tokens_if_needed(token_id, amount);
```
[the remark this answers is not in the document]
kiseln: This is why burning in a centralized place like `ft_on_transfer` is less error prone but then it's harder to handle errors

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
karim-en: The `msg` has an arbitrary size, which means that we can't guarantee the callback execution with fixed gas

near/omni-bridge/src/lib.rs
```diff
@@ -1402,10 +1462,7 @@ impl Contract {
         );
         storage_deposit_action_index += 1;
 
-        let amount_to_transfer = U128(transfer_message.amount.0 - transfer_message.fee.fee.0);
-        let is_deployed_token = self.deployed_tokens.contains(&token);
-
-        let mut promise = self.send_tokens(token.clone(), recipient, amount_to_transfer, &msg);
+        required_balance = required_balance.saturating_add(ONE_YOCTO);
```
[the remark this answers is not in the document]
karim-en: ca4cdbe7e668bc983e644db6a8705566125ceb62

near/omni-bridge/src/lib.rs
```diff
@@ -1345,9 +1327,87 @@ impl Contract {
...
+    }
 }
 
 impl Contract {
+    fn is_refund_required(is_ft_transfer_call: bool) -> bool {
+        if is_ft_transfer_call {
+            match env::promise_result(0) {
+                PromiseResult::Successful(value) => {
+                    if let Ok(amount) = near_sdk::serde_json::from_slice::<U128>(&value) {
+                        // Normal case: refund if the used token amount is zero
```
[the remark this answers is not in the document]
karim-en: Added comments ca4cdbe7e668bc983e644db6a8705566125ceb62

near/omni-bridge/src/lib.rs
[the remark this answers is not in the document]
karim-en: d8ae2fbafd215360767ecc0f95a696e1c637f17d

PR #374, opened 2025-08-05
feat: introduce exponential delay and event ttl

frolvanya:


omni-relayer/src/workers/mod.rs
```diff
@@ -152,26 +168,94 @@ pub async fn process_events(
...
+                            &mut redis_connection_clone,
+                            utils::redis::EVENTS,
+                            &key,
+                        )
+                        .await;
+                        continue;
+                    }
+                };
+
+            if current_timestamp - retryable_event.creation_timestamp
```
karim-en, line 214: Is it safe? In terms of, is it possible that current_timestamp < retryable_event.creation_timestamp

frolvanya: No, I use UTC time which is constant throughout the year for both `current_timestamp` and during the event creation, so it's impossible to have older event having a timestamp greater than current timestamp

karim-en (approved):

PR #380, opened 2025-08-13
Init transfer yield-resume flow

kiseln:
## Overview

This is an extension of `init_transfer` flow on Near that allows bridge users to transfer fungible tokens to the contract first, and make a NEAR deposit for storage and fee as a second step. It uses Near's [yield-resume](https://docs.near.org/smart-contracts/anatomy/yield-resume) functionality to resume execution after storage is deposited.

Note, that payments for storing promise IDs will be taken directly from bridge account but DoS possibilities are limited since the IDs will be deleted after a timeout of 200 blocks.

### Message storage accounts

Now, storage deposit for sending transfers can be credited to a special account calculated based on the message contents and expected sender https://github.com/Near-One/omni-bridge/pull/380/files#diff-b60c5b1d9739aba26a808fe5bb90473c2c30a608886757604357c66a764256b6R1487

The deposit for storing the message can be taken from this account. However, when the message is deleted it will be refunded to the signer of `init_transfer` call (it doesn't make sense to refund it back to the message account since it will be likely never used again). For this reason, it's important to include `signer_id` when calculating this account.

### Updated flow

- if message storage account has enough balance -> use balance of this account
- else, if signer id account has enough balance -> use balance of the signer
- else, create a promise and wait for the storage to be deposited to message storage account
    - when storage deposit arrives -> resume the promise and complete the transfer
    - if storage deposit doesn't arrive until timeout (200 blocks) -> refund the tokens

### Potential Issues

- Changing `required_balance_for_init_transfer` api can break SDK/Integration tests
- Need to still make sure tokens are refunded after the timeout


near/omni-bridge/src/storage.rs
```diff
@@ -170,6 +176,13 @@ impl Contract {
         self.accounts_balances.get(account_id)
     }
 
+    pub fn has_storage_balance(&self, account_id: &AccountId, balance: NearToken) -> bool {
```
karim-en, line 179: Do we really need this a public method?

near/omni-bridge/src/lib.rs
```diff
@@ -1405,6 +1483,27 @@ impl Contract {
...
+    ) -> AccountId {
+        let mut data = Vec::new();
+        data.extend_from_slice(message.token.to_string().as_bytes());
+        data.extend_from_slice(&message.amount.0.to_le_bytes());
+        data.extend_from_slice(message.recipient.to_string().as_bytes());
+        data.extend_from_slice(&message.fee.fee.0.to_le_bytes());
+        data.extend_from_slice(&message.fee.native_fee.0.to_le_bytes());
+        data.extend_from_slice(message.sender.to_string().as_bytes());
+        data.extend_from_slice(message.msg.as_bytes());
+        data.extend_from_slice(signer_id.as_bytes());
```
karim-en, line 1499: I think we can use borsh encoding instead.
Idk why we need the signer here, but it also can be encoded using borsh

near/omni-bridge/src/lib.rs
```diff
@@ -489,18 +496,89 @@ impl Contract {
...
+                .as_bytes(),
+                INIT_TRANSFER_RESUME_GAS,
+                GasWeight(0),
+                PROMISE_REGISTER_ID,
+            );
+
+            let yield_id: CryptoHash = env::read_register(PROMISE_REGISTER_ID)
+                .expect("ERR_READ_PROMISE_REGISTER")
+                .try_into()
+                .expect("ERR_READ_PROMISE_REGISTER");
```
karim-en, line 528: Let's use the `sdk_expect` instead

near/omni-bridge/src/lib.rs
```diff
@@ -489,18 +496,89 @@ impl Contract {
...
+
+            self.init_transfer_promises
```
karim-en, line 530: It will be good later to figure out how to reserve a storage for this in advance, for example it could be done by decreasing the balance of self account.

near/omni-bridge/src/lib.rs
```diff
@@ -489,18 +496,89 @@ impl Contract {
...
+                transfer_message,
+                message_storage_account_id,
+                storage_owner,
+            ),
+            Err(_) => env::log_str("Init transfer resume timeout"),
+        }
+    }
+
+    #[private]
+    pub fn init_transfer_internal(
```
karim-en, line 560: If it is internal then we can remove the `pub` and `#[private]`, and move it down to the private methods sections

karim-en (changes requested):

near/omni-bridge/src/storage.rs
```diff
@@ -185,31 +198,40 @@ impl Contract {
             .saturating_mul((Self::get_basic_storage() + key_len + value_len).into())
     }
 
-    pub fn required_balance_for_init_transfer(&self) -> NearToken {
+    pub fn required_balance_for_init_transfer(&self, msg: Option<String>) -> NearToken {
```
karim-en, line 201: Could you please remind me why we need to pass the `msg` here? since the regular init transfers doesn't accept additional msg.

near/omni-bridge/src/lib.rs
```diff
@@ -489,18 +496,89 @@ impl Contract {
...
+        transfer_message: TransferMessage,
+        message_storage_account_id: AccountId,
+        storage_owner: AccountId,
+        #[callback_result] response: Result<(), PromiseError>,
+    ) {
+        self.init_transfer_promises
+            .remove(&message_storage_account_id);
+
+        match response {
+            Ok(()) => self.init_transfer_internal(
```
karim-en, line 550: if the `init_transfer_internal` panic then the promise will not be removed from the storage

near/omni-bridge/src/storage.rs
kiseln: Not related to this PR I supposed but we did it for more precise calculations in finalizations to 3rd chains

near/omni-bridge/src/storage.rs
kiseln: Impossible to use it as private from the main file

near/omni-bridge/src/lib.rs
kiseln: 89f75a1811996f4f237de21ee14e7b1abb111af9

near/omni-bridge/src/lib.rs
kiseln: Makes sense. Done here 87ed79352aebf10129ac766f42d1ce88c74b55d5

near/omni-bridge/src/storage.rs
karim-en: Let's use `pub(crate)`

near/omni-bridge/src/lib.rs
```diff
@@ -196,44 +200,36 @@ pub struct Contract {
...
-                    token_id,
-                    amount,
-                    init_transfer_msg,
-                );
-                PromiseOrValue::Value(U128(0))
-            }
+        let promise_or_promise_index_or_value = match parsed_msg {
+            BridgeOnTransferMsg::InitTransfer(init_transfer_msg) => self.init_transfer(
+                sender_id,
+                storage_payer,
```
frolvanya, line 220: nitpick: it was renamed to `signer_id` in `init_transfer` method, so maybe it'd be better to rename it here to `signer_id` as well

near/omni-bridge/src/lib.rs
```diff
@@ -488,21 +485,76 @@ impl Contract {
...
 
-        self.update_storage_balance(
-            storage_payer,
-            required_storage_balance,
-            NearToken::from_yoctonear(0),
-        );
+            let yield_id: CryptoHash = env::read_register(PROMISE_REGISTER_ID)
+                .sdk_expect("ERR_READ_PROMISE_REGISTER")
+                .try_into()
+                .sdk_expect("ERR_READ_PROMISE_REGISTER");
```
frolvanya, line 527: Maybe it's better to use different errors here to make it easier to differentiate between `read_register` or `try_into` failure in the future?

frolvanya:
@karim-en @kiseln was refund tested. Does it work as expected?

karim-en:
Please fix the conflicts

kiseln:
> @karim-en @kiseln was refund tested. Does it work as expected?

In case of the timeout - yes

near/omni-bridge/src/lib.rs
kiseln: 8ecbc4ba99a780e2fccf1d1b0c8c260d3997fcfd

near/omni-bridge/src/storage.rs
kiseln: 8ecbc4ba99a780e2fccf1d1b0c8c260d3997fcfd

karim-en:
Please apply the fmt

karim-en:
Let's also bump the contract version

karim-en (approved):

PR #406, opened 2025-09-03
fix: Use borsh to calculate message storage account

kiseln:


karim-en:
We have to bump the version

karim-en (approved):

PR #418, opened 2025-09-15
End-to-end test: Ethereum to Bitcoin

olga24912:


frolvanya (approved):
[This](https://github.com/Near-One/omni-bridge/blob/main/e2e-testing/snakefiles/04_btc_transfer.smk) file also needs to be fixed, but I guess we can do it in a separate PR

olga24912:
> [This](https://github.com/Near-One/omni-bridge/blob/main/e2e-testing/snakefiles/04_btc_transfer.smk) file also needs to be fixed, but I guess we can do it in a separate PR

Yes, let’s do it in a separate PR.

PR #420, opened 2025-09-17
feat: subsidize storage by signer

karim-en:


near/omni-bridge/src/lib.rs
```diff
@@ -561,7 +565,12 @@ impl Contract {
         self.remove_promise(&message_storage_account_id);
 
         if response.is_ok() {
-            self.init_transfer_internal(transfer_message, message_storage_account_id, storage_owner)
+            self.pay_native_fee_from_message_account(
```
kiseln, line 568: We can't panic here. I suggest to make `pay_native_fee_from_message_account` return result instead

near/omni-bridge/src/storage.rs
```diff
@@ -183,6 +183,44 @@ impl Contract {
         }
     }
 
+    // Deducts `native_fee` from `account_id` and credits the remaining balance to `signer_id` and verify the `signer_id` has enough balance.
+    // Returns an error if `account_id` does not have enough balance to cover `native_fee` or if `signer_id` is not registered.
+    pub(crate) fn try_pay_native_fee_from_message_account(
+        &mut self,
+        account_id: &AccountId,
+        native_fee: u128,
+        signer_id: &AccountId,
```
kiseln, line 192: `storage_payer` is probably a better fit for this argument

karim-en: 5753bf490dc3edf09a3a43d057eba97c783e0ea0

near/omni-bridge/src/storage.rs
```diff
@@ -183,6 +183,44 @@ impl Contract {
...
+        &mut self,
+        account_id: &AccountId,
+        native_fee: u128,
+        storage_payer: &AccountId,
+        required_storage_payer_balance: NearToken,
+    ) -> Result<(), String> {
+        let balance = self
+            .accounts_balances
+            .get(account_id)
+            .ok_or("ERR_ACCOUNT_NOT_REGISTERED")?;
```
olga24912, line 198: Why is this an error? If we don’t have a native fee, then we don’t need to register an account.

near/omni-bridge/src/lib.rs
```diff
@@ -493,26 +486,30 @@ impl Contract {
...
+            || (self.has_storage_balance(
+                &signer_id,
+                required_storage_balance.saturating_add(NearToken::from_yoctonear(
+                    init_transfer_msg.native_token_fee.0,
+                )),
+            ) && (init_transfer_msg.native_token_fee.0 == 0
+                || !self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone())))
+        {
+            PromiseOrPromiseIndexOrValue::Value(
+                self.init_transfer_internal(transfer_message, signer_id),
```
olga24912, line 511: And where do we pay for the native_fee? It looks like in the first case we’ve already paid for the native_fee, while in the second we only checked that we can.

near/omni-bridge/src/lib.rs
```diff
@@ -458,13 +458,6 @@ impl Contract {
         amount: U128,
         init_transfer_msg: InitTransferMsg,
     ) -> PromiseOrPromiseIndexOrValue<U128> {
-        // Avoid extra storage read by verifying native fee before checking the role
-        if init_transfer_msg.native_token_fee.0 > 0
-            && self.acl_has_role(Role::NativeFeeRestricted.into(), signer_id.clone())
```
olga24912, line 463: Maybe we should remove this role and make it so that the implicit account always pays for the native_fee. What functionality would break in that case?

near/omni-bridge/src/lib.rs
karim-en: Fixed in 2a7fa2ac84b65fdd4a581879b6f0ad7f3af85e86

karim-en: Maybe , let's think more for the edge cases.

karim-en: Here is the implementation of the first option

near/omni-bridge/src/storage.rs
```diff
@@ -183,6 +183,40 @@ impl Contract {
         }
     }
 
+    // Deducts the total balance from `account_id` and credits it to `storage_payer` and verify the `storage_payer` has enough balance.
+    // Returns an error if `account_id` does not have enough balance to cover `native_fee` or if `signer_id` is not registered.
+    pub(crate) fn try_to_transfer_balance_from_message_account(
```
kiseln, line 188:
```suggestion
    // Used when native fee for the transfer is deposited to the dedicated message account.
    // Deducts the total balance from `account_id` and credits it to `storage_payer`.
    // Returns an error if `account_id` does not have enough balance to cover `native_fee` or if `storage_payer` doesn't have enough balance to complete the transfer.
    pub(crate) fn try_to_transfer_balance_from_message_account(
```

PR #421, opened 2025-09-17
Fix E2E btc transfer test

olga24912:


frolvanya (approved):
It's already fixed in https://github.com/Near-One/omni-bridge/pull/419, but we can merge this one if you want

PR #433, opened 2025-10-15
fix: retry eth client errors

frolvanya:
Recently transfer was stuck with this log:
```text
2025-10-08T15:20:54.293831943+00:00  WARN Failed to finalize transfer (1473): EthRpcError(EthClientError(ParseError(Error("invalid type: null, expected struct TransactionReceipt", line: 0, column: 0))))
```

This can happen when node returns something unexpected, so it's better to retry this. For example, `eth_getTransactionReceipt` can return null if the tx is not yet mined, was dropped/replaced, the node is lagging/syncing, or the tx/block was reorged away. Also we could receive rate limit in `result` json, this needs to be retried as well

### DIFF
--- a/omni-relayer/src/workers/evm.rs
+++ b/omni-relayer/src/workers/evm.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::{Context, Result};
-use bridge_connector_common::result::BridgeSdkError;
+use bridge_connector_common::result::{BridgeSdkError, EthRpcError};
 use tracing::{info, warn};
 
 use near_bridge_client::{NearBridgeClient, TransactionOptions};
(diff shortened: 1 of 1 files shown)


omni-relayer/src/workers/evm.rs
```diff
@@ -251,6 +251,12 @@ pub async fn process_init_transfer_event(
                     log.origin_nonce, block
                 );
                 return Ok(EventAction::Retry);
+            } else if let BridgeSdkError::EthRpcError(EthRpcError::EthClientError(err)) = err {
```
karim-en: As I mentioned earlier, it's better to handle known errors when a retry isn't needed, and retry otherwise.
Because the number of cases where the retry isn't need are much less than when the retry is needed.

frolvanya: Agree, I forgot about this for some reason. I can create a PR refactoring error handling now or we can address this issue during the next relayer refactor/rewrite. Which option do you prefer?

PR #434, opened 2025-10-16
feat: support `safe_verify_deposit`

frolvanya:
Relayer will call `safe_verify_deposit` if `safe_deposit` message was provided during deposit message construction

Related PRs (this PR can be merged only after them):
- https://github.com/Near-One/bridge-sdk-rs/pull/207
- https://github.com/Near-One/bridge-indexer-rs/pull/245

closes https://github.com/Near-One/omni-bridge-private/issues/64

### DIFF
--- a/omni-relayer/Cargo.lock
+++ b/omni-relayer/Cargo.lock
@@ -2254,7 +2254,7 @@ dependencies = [
 [[package]]
 name = "bridge-connector-common"
 version = "0.2.1"
-source = "git+https://github.com/Near-One/bridge-sdk-rs?rev=2bf097216bcfbebb85349219e7cb538b6ca90e6b#2bf097216bcfbebb85349219e7cb538b6ca90e6b"
+source = "git+https://github.com/Near-One/bridge-sdk-rs?rev=db879821d885e12d1b1cb132e11dd6b24efa8f77#db879821d885e12d1b1cb132e11dd6b24efa8f77"
 dependencies = [
  "eth-proof",
  "ethers",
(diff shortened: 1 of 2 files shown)


omni-relayer/src/utils/nonce.rs
```diff
@@ -64,7 +64,11 @@ impl NonceManager {
         }
 
         let reserved = *local_nonce;
-        *local_nonce += 1;
+        if let ChainClient::Near { .. } = self.client {
+            *local_nonce += 2;
```
frolvanya: There's a case where two concurrent transfers will block each other:
1) transfer A makes to storage deposit. Current nonce: 0
2) transfer B makes a call to `fin_btc_transfer`. Current nonce: 2
3) transfer B goes through, but when we try to call `fin_btc_transfer` with nonce 1 it will fail, since transaction with higher nonce (2) was already submitted to chain

omni-relayer/src/workers/utxo.rs
kiseln: Are these of different type? Why do we have to manually assign every field?

omni-relayer/src/utils/nonce.rs
frolvanya: Nope, this should be fixed before merging. I'll do this today

frolvanya: Fixed in https://github.com/Near-One/bridge-sdk-rs/pull/212 + https://github.com/Near-One/omni-bridge/pull/434/commits/a076ec473787b74621cdc220885c07950c9f6575

omni-relayer/src/workers/utxo.rs
frolvanya: Yes, there're different `DepositMsg` from `bridge-indexer-types` and `bridge-sdk-rs`

omni-relayer/src/utils/nonce.rs
frolvanya: No, that was exactly the reason of the bug. Before this we reserved 2 nonces at the same time, but we could be in a weird situation where nonce x, x+2 are used, but x+1 is just getting submitted and it will fail since x+2 was already used

PR #447, opened 2025-10-31
E2E tests: add the ability to run tests for deployed contracts.

olga24912:
So far, I’ve only modified one test: the BTC <-> NEAR transfer.
* Waiting for a sufficient number of Bitcoin transaction confirmations.
* Running for the default contracts and verifying the results via the API.


frolvanya (approved):

frolvanya (changes requested):
Also, let's use non-reproducible-wasm instead of reproducible builds for e2e. We can either change it here (we need to duplicate current rules in Makefile to do it) or in a new PR

PR #463, opened 2025-11-17
feat: added `finalised_utxo_transfers` guard

frolvanya:


near/omni-bridge/src/lib.rs
```diff
@@ -2116,11 +2167,25 @@ impl Contract {
...
     fn utxo_fin_transfer_to_near(
+        &mut self,
         recipient: AccountId,
         token_id: AccountId,
         amount: U128,
         utxo_fin_transfer_msg: UtxoFinTransferMsg,
+        origin_chain: ChainKind,
+        storage_owner: &AccountId,
     ) -> Promise {
+        let required_storage_balance = self.add_fin_utxo_transfer(&UnifiedTransferId {
```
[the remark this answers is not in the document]
frolvanya: Like this https://github.com/Near-One/omni-bridge/pull/463/commits/ba9656038724c5a032c018adda6ff1aa54a9d9c8?

near/omni-bridge/src/lib.rs
```diff
@@ -901,27 +899,51 @@ impl Contract {
...
         amount: U128,
         utxo_fin_transfer_msg: UtxoFinTransferMsg,
+        origin_chain: ChainKind,
+        storage_owner: &AccountId,
     ) -> U128 {
-        Self::resolve_utxo_fin_transfer_internal(token_id, amount, utxo_fin_transfer_msg)
+        let required_storage_balance = self.add_fin_utxo_transfer(&UnifiedTransferId {
+            origin_chain,
+            kind: TransferIdKind::Utxo(utxo_fin_transfer_msg.utxo_id.clone()),
+        });
```
olga24912: It looks strange that we add a transaction in one function and then immediately remove it. In that case, it would make more sense to only add it if needed, or to add it one step earlier.

olga24912 (changes requested):

near/omni-bridge/src/lib.rs
frolvanya: https://github.com/Near-One/omni-bridge/pull/463/commits/a660b9a7ead1b6c75d0a1efff3278da1cc1529ab

near/omni-bridge/src/lib.rs
```diff
@@ -1262,8 +1273,14 @@ impl Contract {
             .sdk_expect("The transfer does not exist")
     }
 
-    pub fn is_transfer_finalised(&self, transfer_id: TransferId) -> bool {
-        self.finalised_transfers.contains(&transfer_id)
+    pub fn is_transfer_finalised(&self, transfer_id: &UnifiedTransferId) -> bool {
```
karim-en: This break the backward compatibility, let's add a new method, or accept two kind of arguments using option

near/omni-bridge/src/lib.rs
```diff
@@ -2217,14 +2268,28 @@ impl Contract {
...
         amount: U128,
         utxo_fin_transfer_msg: UtxoFinTransferMsg,
+        origin_chain: ChainKind,
+        storage_owner: &AccountId,
     ) -> U128 {
         let is_ft_transfer_call = !utxo_fin_transfer_msg.msg.is_empty();
         if Self::is_refund_required(is_ft_transfer_call) {
             amount
         } else {
+            let required_storage_balance = self.add_fin_utxo_transfer(&UnifiedTransferId {
```
karim-en: We have to add the fin transfer before transferring the tokens, otherwise the fin utxo transfer will be reverted but the recipent received the tokens

karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -2217,14 +2268,28 @@ impl Contract {
...
+                origin_chain,
+                kind: TransferIdKind::Utxo(utxo_fin_transfer_msg.utxo_id.clone()),
+            });
+
+            self.update_storage_balance(
```
karim-en: We can't panic here, the transfer already processed

karim-en (changes requested):

kiseln (approved):

olga24912 (approved):

near/omni-bridge/src/lib.rs
```diff
@@ -2062,8 +2111,15 @@ impl Contract {
         }
 
         if let OmniAddress::Near(recipient) = utxo_fin_transfer_msg.recipient.clone() {
-            Self::utxo_fin_transfer_to_near(recipient, token_id, amount, utxo_fin_transfer_msg)
-                .into()
+            Self::utxo_fin_transfer_to_near(
```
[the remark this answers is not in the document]
frolvanya: I don't understand where's a race condition here. Let's say we have two fast transfers that try to process same init transfer. First transfer started a bit earlier and now on a callback, while second transfer just started. Now first transfer added fin transfer message to the state and second transfer entered the callback. In this case adding the same message for the second time should fail here:
https://github.com/Near-One/omni-bridge/pull/463/changes#diff-b60c5b1d9739aba26a808fe5bb90473c2c30a608886757604357c66a764256b6R1867-R1870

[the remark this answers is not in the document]
karim-en: This turned into spaghetti code, I'm going to push mini PR to fix this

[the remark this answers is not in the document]
karim-en: @frolvanya check [this](https://github.com/Near-One/omni-bridge/pull/484/) pelase

kiseln (approved):

karim-en (approved):

PR #464, opened 2025-11-19
feat: fee bumping

kiseln:
# Overview

This PR helps to handle gas spikes on Ethereum. Occasionally we submit a transaction just before the spike, which means it gets stuck in the mempool. All further transactions are then stuck as well since execution on Ethereum is sequential.

We've implemented a separate thread that watches transactions we've sent. If a transaction is stuck due to low gas, the gas fee will be bumped according to current market price.

## Implementation

Pending transactions are stored in a Redis sorted set. Transactions are added on send if the configuration is enabled for a target chain.

From a sorted set transactions are retrieved one by one starting from the lowest nonce. If transaction is detected in mempool and it has been there for a certain (configured) time, we fetch current gas fee and send a replacement transaction.

### Dropped transactions

As an additional related feature, If transaction is dropped (missing from mempool and not included) we schedule it for sending again.

### DIFF
--- a/omni-relayer/src/startup/evm_fee_bumping.rs
+++ b/omni-relayer/src/startup/evm_fee_bumping.rs
@@ -283,6 +283,7 @@ fn sleep(check_interval_seconds: u64) -> Sleep {
 }
 
 #[allow(clippy::as_conversions)]
+#[allow(clippy::cast_precision_loss)]
 pub fn wei_to_gwei(wei: u128) -> f64 {
     let gwei = wei / 1_000_000_000;
     gwei as f64


omni-relayer/src/startup/evm_fee_bumping.rs
```diff
@@ -0,0 +1,292 @@
...
+            }
+        };
+    }
+}
+
+fn should_bump(
+    fee_bumping_config: &config::FeeBumping,
+    original_fee: Eip1559Estimation,
+    suggested_fee: Eip1559Estimation,
+) -> ShouldBump {
```
frolvanya: Can we simplify it to `anyhow::Result<Eip1559Estimation>`. I don't understand why do we need enum here

omni-relayer/src/workers/near.rs
```diff
@@ -740,3 +766,35 @@ pub async fn initiate_fast_transfer(
...
+        chain_kind,
+        source_event_id,
+        omni_bridge_event,
+    );
+
+    utils::redis::zadd(
+        config,
+        redis_connection_manager,
+        &utils::pending_transactions::get_pending_tx_key(chain_kind),
+        nonce as f64,
```
frolvanya: Why do we need to store nonce as `f64`?

omni-relayer/src/startup/evm_fee_bumping.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    let provider = ProviderBuilder::new()
+        .wallet(wallet.clone())
+        .connect_http(evm_config.rpc_http_url.parse().context("Invalid RPC URL")?);
+
+    Ok(provider)
+}
+
+async fn sleep(check_interval_seconds: u64) {
+    tokio::time::sleep(tokio::time::Duration::from_secs(check_interval_seconds)).await;
```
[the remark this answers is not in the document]
kiseln: Not sure I understand, this is how it's used across the codebase. Made a slight optimization f9a8888d6dc15a5196cbb212b352dbdcb322c0d4

PR #475, opened 2025-11-28
fix: resume in next block

karim-en:
Due to protocol restrictions, the resume can't be called after yield in the same block.
Example of timeout error https://nearblocks.io/txns/H4Dru2hsfS8HuEBgUcudnaAji5x2KqjdZmtcFV1fFspV

### DIFF
--- a/near/omni-bridge/src/lib.rs
+++ b/near/omni-bridge/src/lib.rs
@@ -567,7 +567,7 @@ impl Contract {
         #[callback_result] response: Result<(), PromiseError>,
     ) -> U128 {
         self.remove_promise(&message_storage_account_id);
-        if !response.is_ok() {
+        if response.is_err() {
             env::log_str("Init transfer resume timeout");
         }
 


near/omni-bridge/src/storage.rs
```diff
@@ -160,14 +160,26 @@ impl Contract {
...
 
+    #[private]
+    pub fn resume_promise(&self, account_id: &AccountId) -> PromiseOrValue<()> {
+        if let Some(promise_id) = &self.init_transfer_promises.get(account_id) {
+            let result = env::promise_yield_resume(promise_id, &[]);
+            env::log_str(&format!("Resume promise. Result: {result}"));
+
+            if !result {
+                return Self::ext(env::current_account_id())
+                    .resume_promise(&account_id)
```
karim-en: verify if it is ok to create the promise without fixed gas

karim-en: https://testnet.nearblocks.io/txns/5RCmMaBkWHnXcYtQJzrgdqZ36R8q6orGFKNHdZG5MJyt?tab=summary

near/omni-bridge/src/storage.rs
```diff
@@ -160,14 +160,26 @@ impl Contract {
...
 
+    #[private]
+    pub fn resume_promise(&self, account_id: &AccountId) -> PromiseOrValue<()> {
+        if let Some(promise_id) = &self.init_transfer_promises.get(account_id) {
+            let result = env::promise_yield_resume(promise_id, &[]);
+            env::log_str(&format!("Resume promise. Result: {result}"));
+
+            if !result {
+                return Self::ext(env::current_account_id())
+                    .resume_promise(account_id)
```
frolvanya: What will be the worst case scenario if this recursive call will be called on every iteration. Will we just run out of gas without refunding tokens back to user?

I understand that it shouldn't be a case where there's more than one iteration, but it's still would be nice to artificially limit the depth of the recursion here, no?

mitinarseny: > We could could return promise or value in the storage_deposit, but not sure that the integrated dapps can handle this, because it is not standard behaviour.

With existing spec of [StoraeManagement](https://github.com/near/NEPs/blob/master/neps/nep-0145.md) standard, it's impossible to write a smart-contract that refunds in case of failed `receiver_id::storage_deposit()` in a safe way. Moreover, the same applies even for failed native transfers due to Near's Promise design. That's the reason for `intents.near` to not refund wNEAR in both these cases.

near/omni-bridge/src/lib.rs
```diff
@@ -567,23 +567,21 @@ impl Contract {
...
-            ) {
-                env::log_str(&format!("Error paying native fee and storage: {err}"));
-                return transfer_message.amount;
-            }
-
-            self.init_transfer_internal(transfer_message, storage_owner)
-        } else {
+        if response.is_err() {
             env::log_str("Init transfer resume timeout");
-            transfer_message.amount
```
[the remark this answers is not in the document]
karim-en: There is a flow where we can get timeout but the storage has been already paid and resumed.

PR #480, opened 2025-12-09
Migration for near intents tokens

kiseln:


near/omni-token/src/lib.rs
```diff
@@ -179,39 +175,22 @@ impl MetadataManagment for OmniToken {
...
     #[payable]
     fn ft_transfer(&mut self, receiver_id: AccountId, amount: U128, memo: Option<String>) {
-        self.token.ft_transfer(receiver_id, amount, memo);
+        // Legacy bridging flow used by Near Intents
+        if receiver_id == env::current_account_id()
+            && memo
+                .as_deref()
+                .map_or(false, |memo| memo.starts_with(WITHDRAW_MEMO_PREFIX))
+        {
+            if let Some(withdraw_relayer) = self.read_withdraw_relayer_address() {
```
karim-en: It will be good to have a setter for this in case if we want to update the relayer account.

near/omni-token/src/migrate.rs
```diff
@@ -0,0 +1,79 @@
...
+            require!(env::storage_remove(OWNABLE_KEY), "Wrong token version for migration");
+
+            if let Some(relayer) = withdraw_relayer {
+                env::storage_write(WITHDRAW_RELAYER_ADDRESS, &borsh::to_vec(&relayer).unwrap());
+            }
+            
+            Self {
+                controller,
+                token: state.token,
+                metadata: LazyOption::new(
```
karim-en: We have to be careful with this, both of the fields use some key, so the data isn't stored in the state struct directly.

near/omni-token/src/migrate.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub token: FungibleToken,
+    pub metadata: Lazy<FungibleTokenMetadata>,
+}
+
+#[near]
+impl OmniToken {
+    #[private]
+    #[init(ignore_state)]
+    #[allow(unused_variables)]
+    pub fn migrate(controller: AccountId, withdraw_relayer: Option<AccountId>) -> Self {
```
karim-en: The migration function is private, so only the self account can call it.
We have to align on the migration flow before releasing this.

karim-en:
Let's also implement the migration logic in the Omni contract.

near/omni-token/src/migrate.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+#[near]
+impl UpgradeAndMigrate for OmniToken {
+    fn upgrade_and_migrate(&self) {
+        self.assert_controller();
+
+        // Receive the code directly from the input to avoid the
+        // GAS overhead of deserializing parameters
+        let code = env::input().unwrap_or_else(|| env::panic_str("ERR_NO_INPUT"));
```
mitinarseny: Maybe add support for global contracts?

near/omni-token/src/migrate.rs
kiseln: It's built on top of global contracts PR.
Migration will be done with full access key so private is fine.

frolvanya (commented):
Please fix CI as well

near/omni-token/src/migrate.rs
```diff
@@ -0,0 +1,81 @@
...
+        // Deploy the contract code.
+        let promise_id = env::promise_batch_create(&env::current_account_id());
+        env::promise_batch_action_deploy_contract(promise_id, &code);
+        // Call promise to migrate the state.
+        // Batched together to fail upgrade if migration fails.
+        env::promise_batch_action_function_call(
+            promise_id,
+            "migrate",
+            &json!({ "from_version": CURRENT_STATE_VERSION })
```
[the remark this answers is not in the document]
frolvanya: Then maybe it's better to remove this method completely if
1) We won't use it for this migration
2) It'll fail, since provided controller is `None`?

near/omni-token/src/migrate.rs
```diff
@@ -0,0 +1,83 @@
...
+                "Wrong token version for migration"
+            );
+
+            let relayer = withdraw_relayer.unwrap_or_else(|| env::panic_str("Withdraw relayer must be provided"));
+            env::storage_write(WITHDRAW_RELAYER_ADDRESS, &borsh::to_vec(&relayer).unwrap());
+
+            Self {
+                controller: controller.unwrap_or_else(|| env::panic_str("Controller must be provided")),
+                token: state.token,
+                metadata: LazyOption::new(b"m".to_vec(), Some(&state.metadata.get())),
```
karim-en: We have to be careful with the previous [storage prefix](https://github.com/near/intents/blob/8fafb37c82a2cb4ea6a26cead95551945790dda5/poa-token/src/contract.rs#L215). In the Omni contract we use b"t" and b"m", but the PoA uses 0 & 1 bytes from enum.

For the token struct, I suppose that the controller will be migrated with the old prefix, and this is ok, but could be confusing for the code reader.
But for the metadata the storage is migrated to `m`  prefix, but the storage with prefix `1` will still be occupied by old metadata.

I propose to:
- Add comments that the storage prefix could be `0` even if set `t` in the constructor.
- In the migrate function call remove for the old lazy metadata
- Find a way to mark this kind of tokens, unfortunately  the [key_prefix](https://docs.rs/near-sdk/latest/src/near_sdk/collections/lookup_map.rs.html#19) field is private, but looks like we can [print it](https://docs.rs/near-sdk/latest/src/near_sdk/collections/lookup_map.rs.html#375).

near/omni-bridge/src/lib.rs
```diff
@@ -2382,6 +2427,24 @@ impl Contract {
...
+    ) -> Promise {
+        let new_token = self
+            .migrated_tokens
+            .get(&old_token)
+            .sdk_expect("ERR_TOKEN_NOT_MIGRATED");
+
+        let burn = ext_token::ext(old_token.clone()).burn(amount);
+        let mint = ext_token::ext(new_token.clone()).mint(sender_id, amount, None);
+
+        burn.and(mint)
```
r-near: We need handling if the mint fails - we have to reverse the burn, right?

kiseln: Highly unlikely any of these will fail. It's also a low-usage method so we don't want to complicate it

karim-en (approved):

PR #486, opened 2025-12-15
refactor: use enum for errors

frolvanya:


near/omni-types/src/errors.rs
```diff
@@ -0,0 +1,338 @@
+use core::fmt;
+use near_sdk::{AccountId, NearToken};
+
+pub trait ErrorCode {
+    fn code(&self) -> &'static str;
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BridgeError {
```
frolvanya: I can split up this huge enum into smaller ones if this is needed. Technically they could be grouped independently, so it won't even introduce any unnecessary nesting

near/omni-types/src/errors.rs
```diff
@@ -0,0 +1,337 @@
...
+    TokenNotRegistered,
+    TransferNotExist,
+    FastTransferNotFound,
+    UtxoConfigMissing,
+}
+
+impl ErrorCode for BridgeError {
+    fn code(&self) -> &'static str {
+        match self {
+            Self::InvalidMetadata => "ERR_INVALID_METADATA",
```
karim-en: Isn't [thiserror](https://docs.rs/thiserror/latest/thiserror/) much easier to use and maintain?

near/omni-types/src/errors.rs
kiseln: If we want to have strong typing for consumers these should be under one enum

near/omni-types/src/errors.rs
```diff
@@ -0,0 +1,337 @@
...
+    ReadPromiseYieldId,
+    TokenNotFound,
+    TokenNotRegistered,
+    TransferNotExist,
+    FastTransferNotFound,
+    UtxoConfigMissing,
+}
+
+impl ErrorCode for BridgeError {
+    fn code(&self) -> &'static str {
```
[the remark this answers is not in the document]
frolvanya: We need to pass a `&str` in `require` macro. That's the whole purpose of this method. I can change name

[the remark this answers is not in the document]
kiseln: Then I agree with using macros to generate string values and yes, I would name it something else (could even be `Display` implementation)

near/omni-types/src/tests/lib_test.rs
```diff
@@ -469,3 +470,31 @@ fn test_chain_kind_from_str() {
...
+        StorageBalanceError::AccountNotRegistered("near".parse().unwrap()).to_string(),
+        "ERR_ACCOUNT_NOT_REGISTERED"
+    );
+    assert_eq!(
+        StorageBalanceError::NotEnoughStorage {
+            required: NearToken::from_yoctonear(100),
+            available: NearToken::from_yoctonear(50),
+        }
+        .to_string(),
+        "ERR_NOT_ENOUGH_STORAGE"
```
frolvanya: Take into account that logs won't contain information about any fields in enum variants

1) We can leave it as it's for some internal usage
2) We can remove fields from enum variants completely
3) We can make a custom serialization specifically for such cases

What's better?

near/omni-bridge/src/storage.rs
```diff
@@ -428,12 +441,15 @@ impl Contract {
...
+            .parse()
+            .sdk_expect(BridgeError::ParseAccountId.as_ref());
 
         borsh::to_vec(&max_account_id)
-            .sdk_expect("ERR_BORSH")
+            .sdk_expect(BridgeError::Borsh.as_ref())
             .len()
             .try_into()
-            .sdk_expect("ERR_CAST")
+            .sdk_expect(BridgeError::Cast.as_ref())
```
frolvanya: I believe we can completely remove all `.as_ref()` in `sdk_expect` method if instead of `msg: &str` we would use  `msg: impl Into<&str>`, this way we can pass all enums directly

near/omni-bridge/src/btc.rs
```diff
@@ -46,7 +47,7 @@ impl Contract {
             {
                 require!(
                     btc_address == target_btc_address,
-                    "Incorrect target address"
+                    BridgeError::IncorrectTargetAddress.code()
```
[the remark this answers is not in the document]
frolvanya: Changed in https://github.com/Near-One/omni-bridge/pull/486/commits/290edb82ba90dfa88312c6c1efd9964722012adb

near/omni-types/src/errors.rs
frolvanya: Used strum instead (we already used it in omni-types, so no extra dependency is needed)

https://github.com/Near-One/omni-bridge/pull/486/commits/290edb82ba90dfa88312c6c1efd9964722012adb

near/omni-types/src/errors.rs
```diff
@@ -0,0 +1,338 @@
...
```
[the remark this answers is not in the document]
frolvanya: Changed all enum errors to follow `ERR_ERROR_NAME` naming conventions in https://github.com/Near-One/omni-bridge/pull/486/commits/290edb82ba90dfa88312c6c1efd9964722012adb

near/omni-types/src/errors.rs
frolvanya: Cleaned up in https://github.com/Near-One/omni-bridge/pull/486/commits/290edb82ba90dfa88312c6c1efd9964722012adb

near/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace.package]
 authors = ["Near One <info@nearone.org>"]
 repository = "https://github.com/Near-One/omni-bridge"
-version = "0.4.3"
+version = "0.4.4"
```
frolvanya: @karim-en Should I leave bumped version in this PR or it's better to stick to 0.4.3 for now and release 0.4.4 with a bigger change?

olga24912 (approved):

karim-en (approved):

near/omni-types/src/tests/lib_test.rs
karim-en: example: 

```rust
define_contract_errors! {
    "ERR",
    BridgeError {
        Borsh {r1: AccountId, r2: AccountId},
        InvalidState {},
        }
  }
  
      #[test]
    fn test_bridge_error_display_format() {
        let error = BridgeError::InvalidState {};
        let display_str = format!("{}", error);
        assert!(display_str.starts_with("ERR_"));
        assert!(display_str.contains("INVALID_STATE"));
    }

    #[test]
    fn test_bridge_error_display_borsh() {
        let error = BridgeError::Borsh {
            r1: "alice.near".parse().unwrap(),
            r2: "bob.near".parse().unwrap(),
        };
        let display_str = format!("{}", error);
        assert_eq!(display_str, "ERR_BORSH: r1=alice.near, r2=bob.near");
    }

  
  ```

File: near/omni-types/src/errors.rs [hunk 9]
```diff
@@ -0,0 +1,162 @@
...
+    NotEnoughBalanceForFee,
+    SignerNotEnoughBalance,
+    SignerNotRegistered,
+}
+
+#[derive(Display, Debug, Clone, Copy, PartialEq, Eq, AsRefStr)]
+#[strum(serialize_all = "shouty_snake_case", prefix = "ERR_")]
+pub enum TokenError {
+    FailedToReadState,
+    InvalidCodeHash,
```
kiseln: What happens if we introduce another contract that will have migration methods and require same set of errors? Are we going to duplicate these for the new contract?

near/omni-types/src/tests/lib_test.rs
frolvanya: Unified in 14df9f074f2957fb8aed1997241b005ac11d1060

PR #494, opened 2025-12-19
fix: remove token from map if deploy failed

frolvanya:


near/omni-bridge/src/lib.rs
```diff
@@ -1098,6 +1098,30 @@ impl Contract {
         )
     }
 
+    #[private]
+    pub fn deploy_token_by_deployer_callback(
```
karim-en: Let's improve the `test_deploy_token_with_huge_metadata_insufficient_deposit` so we have to check that the token can be deployed second time

near/omni-bridge/src/lib.rs
```diff
@@ -1098,6 +1098,30 @@ impl Contract {
...
+        &mut self,
+        token_address: OmniAddress,
+        token_id: AccountId,
+    ) {
```
karim-en: Let's return the pormise

karim-en (changes requested):

karim-en (approved):

PR #504, opened 2026-01-02
fix(relayer): resync nonce only on errors

frolvanya:


omni-relayer/src/workers/mod.rs
```diff
@@ -176,18 +181,25 @@ pub async fn process_events(
             continue;
         }
 
-        if let Err(err) = near_omni_nonce.resync_nonce().await {
-            warn!("Failed to resync near nonce: {err:?}");
-        }
+        if is_nonce_resync_needed.load(Ordering::Relaxed) {
+            if let Err(err) = near_omni_nonce.resync_nonce().await {
```
kiseln: Do we need to resync for Near at all?

kiseln (commented):
I would be careful about coupling EventAction and nonce logic.

Retry shouldn't always mean that we didn't send a transaction and therefore have to resync. And the opposite, we can Remove for reasons other than being successful.

kiseln:
> > Retry shouldn't always mean that we didn't send a transaction and therefore have to resync. And the opposite, we can Remove for reasons other than being successful.
> 
> That's true, but right now it's resynced after every batch, so even slight optimization is good
> 
> Generally speaking it must be completely rewritten to use channels and require resync only on specific errors, but since we're planning to rewrite relayer soon anyway, I don't think it's worth to make a perfect fix now

To keep it simple I would then do resyncing logic for EVM only

omni-relayer/src/workers/mod.rs
frolvanya: Fixed in https://github.com/Near-One/omni-bridge/pull/504/commits/6e1984d4a934f92c236a208c2a31a1f28bb8e7f2

I think we can optimize it even further by resyncing nonce only for the problematic chains. Instead of all evm, but it'll complicate code a little bit, so I think it's not worth it

omni-relayer/src/workers/mod.rs
```diff
@@ -599,6 +611,15 @@ pub async fn process_events(
                         let omni_connector = omni_connector.clone();
                         let signer = signer.clone();
                         let evm_nonces = evm_nonces.clone();
+                        let is_evm_nonce_resync_needed = is_evm_nonce_resync_needed.clone();
+                        let should_resync_evm_nonce = matches!(
```
olga24912: Maybe it would be better to add an is_evm_chain method?

frolvanya: Forgot about this method. Refactored in https://github.com/Near-One/omni-bridge/pull/504/commits/3fdd9966c8056b04c88279d47c274bdea3e20bcc

omni-relayer/src/workers/mod.rs
```diff
@@ -151,6 +154,20 @@ pub async fn process_events(
         .near_bridge_client()
         .and_then(near_bridge_client::NearBridgeClient::account_id)?;
 
+    near_omni_nonce
+        .resync_nonce()
+        .await
+        .context("Failed to resync near nonce")?;
+
+    if let Some(near_fast_nonce) = near_fast_nonce.clone() {
```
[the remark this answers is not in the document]
frolvanya: Yes, for sending fast transfers. Since fast relayer and regular relayer are two different accounts, then we have to track both nonces

PR #519, opened 2026-01-27
Scripts for deploy and link tokens on HyperEVM/HyperCore

olga24912:


evm/utils/hyperliquid/link_tokens.py
```diff
@@ -0,0 +1,99 @@
...
+
+    print(TOKEN, contract_address.lower())
+    #requestEvmContract(account)
+    finalizeEvmContract(account)
+
+if __name__ == "__main__":
+    main()
+
+# curl -s https://api.hyperliquid-testnet.xyz/info   -H "Content-Type: application/json"   -d '{"type": "tokenDetails", "tokenId": "0x646586ef3576346a4fcc9548909c1cba"}' | jq
+# curl -s https://api.hyperliquid-testnet.xyz/info  -H "Content-Type: application/json"  -d '{ "type": "spotMeta" }' | jq '.tokens[] | select(.name=="JHWL")'
```
karim-en: This should be cleaned up

evm/hardhat.config.ts
```diff
@@ -131,6 +131,36 @@ task("deploy-token-impl", "Deploys the BridgeToken implementation").setAction(as
   )
 })
 
+task("deploy-token-proxy", "Deploy BridgeToken behind a UUPS proxy")
+  .addParam("name", "Token name")
+  .addParam("symbol", "Token symbol")
+  .addOptionalParam("decimals", "Token decimals", "18")
```
karim-en: We have to add scripts for Hyperliquid token with system address.
Also the system address should be retrieved from the rpc in the script

karim-en (changes requested):

PR #538, opened 2026-02-17
feat: switch relayer to consume nats events

kiseln:
## Overview

Relayer will now receive events via Nats Jetstream instead of MongoDb change streams.

Additionally, if action is to be triggered, the relayer will publish an event to subject `relayer.tasks.{chain}`, where `chain` is where the action need to be executed (for instance `near` for signing a transfer, `eth` for finalizing on Ethereum)

In the next step of the refactor, we will replace Redis and use these events for execution


omni-relayer/src/startup/bridge_indexer.rs
```diff
@@ -122,15 +155,17 @@ async fn handle_transaction_event(
             let origin_nonce = sign_event.message_payload.transfer_id.origin_nonce;
             let key = near_event_key(&origin_transaction_id, origin_nonce);
 
-            utils::redis::add_event(
+            let target_chain = sign_event.message_payload.recipient.get_chain();
```
frolvanya: nit: let's use destination chain, since it's more common across the codebase

Also, I believe I've added `TransferMessage::get_destination_chain()` instead of `transfer_message.recipient.get_chain()`, so you may use this instead as well

PR #562, opened 2026-02-20
fix(sol): minor issues and code quality

kiseln:
### max_used_nonce not updating
When finalizing a transfer, `max_used_nonce` value must be updated in the config account. This is needed to properly calculate rent required for nonce accounts. This wasn't happening because the config instance in the common WormholeCPI struct is not marked mutable.

#### Fix
I decided to keep the config in the shared struct immutable. It's used in 6 different methods and only two (finalize_transfer and finalize_transfer_sol) need to mutate the config. Instead, I added a separate mutable config instance, same way it is already done for finalize_transfer_sol

### Wrong account passed to metaplex instruction
The program id is internally ovewritten with the correct instruction, so the code was working as intended. Still rather fix this.

### Remove unwraps
Minor code quality cleanup

### Remove unsafe code
unsafe block in `used_nonces.rs` was unnecessary


solana/programs/bridge_token_factory/src/instructions/user/finalize_transfer.rs
```diff
@@ -7,26 +7,28 @@ use anchor_spl::{
...
 
 #[derive(Accounts)]
 #[instruction(data: SignedPayload<FinalizeTransferPayload>)]
 pub struct FinalizeTransfer<'info> {
+    #[account(
+        mut,
+        seeds = [CONFIG_SEED],
+        bump = config.bumps.config,
+    )]
+    pub config: Box<Account<'info, Config>>,
```
frolvanya: Now same config is accessible via common.config and self.config? Why is it needed? Because we can't mark config in `WormholeCPI` as `mut`?

kiseln:
@karim-en

karim-en:
> @karim-en 

@kiseln

PR #565, opened 2026-02-24
feat: read mpc prover

frolvanya:
Tested with a real transaction on Abstract

initTransfer: https://sepolia.abscan.org/tx/0x8d286a01fa892903128228cdca896de68c7b774ccbd1b46b25867ef9499c6fc3#eventlog 

verify_proof:
https://testnet.nearblocks.io/txns/5A9SusR3FSdooUGBENeE27CdxyE9KXbLTp3ZTNqarCGK?tab=enhanced

### DIFF
--- a/near/omni-prover/mpc-omni-prover/src/tests.rs
+++ b/near/omni-prover/mpc-omni-prover/src/tests.rs
@@ -96,7 +96,7 @@ fn abs_testnet_evm_request() -> EvmRpcRequest {
     EvmRpcRequest {
         tx_id: abs_testnet_tx_id(),
         extractors: vec![EvmExtractor::Log { log_index: 3 }],
-        finality: EvmFinality::Safe,
+        finality: EvmFinality::Latest,
     }
 }
 


near/omni-prover/mpc-omni-prover/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn verify_proof(&self, #[serializer(borsh)] input: Vec<u8>) -> Promise {
+        let args = MpcVerifyProofArgs::try_from_slice(&input).near_expect(ProverError::ParseArgs);
+
+        let sign_payload = ForeignTxSignPayload::try_from_slice(&args.sign_payload)
+            .near_expect(ProverError::ParseArgs);
+
+        let ForeignTxSignPayload::V1(ref payload_v1) = sign_payload;
+
+        let chain_kind = Self::request_to_chain_kind(&payload_v1.request)
```
olga24912: The fact that we are not checking the chain kind looks unsafe.

I’m a bit lazy to go through all the logic in the Omni Bridge right now, but I suspect that this might allow us to use a proof from one chain while presenting it as if it were from another chain

Maybe we should go back to the old design where each chain has its own prover.

olga24912 (changes requested):

PR #573, opened 2026-03-07
feat(near): allow only trusted relayers to call `fast_fin_transfer` method

frolvanya:


karim-en (changes requested):

near/omni-bridge/src/lib.rs
```diff
@@ -779,6 +774,11 @@ impl Contract {
         storage_payer: AccountId,
         fast_fin_transfer_msg: FastFinTransferMsg,
     ) -> PromiseOrPromiseIndexOrValue<U128> {
+        require!(
+            self.is_trusted_relayer(&env::predecessor_account_id()),
```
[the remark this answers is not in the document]
karim-en: yep

karim-en (approved):

karim-en:
@frolvanya let's bump the version to v0.4.7

near/omni-bridge/src/lib.rs
```diff
@@ -743,13 +743,16 @@ impl Contract {
...
+    #[allow(clippy::needless_pass_by_value)]
     fn fast_fin_transfer(
         &mut self,
         token_id: AccountId,
         amount: U128,
-        storage_payer: AccountId,
+        signer_id: AccountId,
         fast_fin_transfer_msg: FastFinTransferMsg,
     ) -> PromiseOrPromiseIndexOrValue<U128> {
+        require!(self.is_trusted_relayer(&signer_id), "Relayer is not active");
```
karim-en, line 754: I've just noticed that we use error string instead of bridge errors

karim-en (changes requested):

kiseln (approved):

near/omni-bridge/src/lib.rs
frolvanya: That's correct, because we throw same error from `omni-utils`, it would be very strange if `sign_transfer` and other methods under `trusted_relayer` procmacro will send "Relayer is not active" while `fast_fin_transfer` will send `ERR_RELAYER_IS_NOT_ACTIVE`

karim-en (approved):

PR #582, opened 2026-03-13
feat: replace relayer event stream from redis to nats

kiseln:


omni-relayer/src/workers/mod.rs
```diff
@@ -155,11 +183,6 @@ pub async fn process_events(
         .near_bridge_client()
         .and_then(near_bridge_client::NearBridgeClient::account_id)?;
 
-    near_omni_nonce
```
frolvanya: Is it okay that nonce now is uninitialized?

PR #585, opened 2026-03-20
fix(omni-relayer): set mpc finalities

frolvanya:


omni-relayer/src/workers/near.rs
```diff
@@ -143,13 +143,7 @@ pub async fn process_transfer_event(
             jsonrpc_client,
             tx_hash,
             signer,
-            &[
```
kiseln: Why are we not retrying these errors?

PR #612, opened 2026-04-13
feat: add external_id to InitTransferMsg and bound string fields

karim-en:
## Summary

- Add a new `external_id: Option<...>` field to `InitTransferMsg`. It's mixed into the virtual storage account hash so otherwise-identical transfers can derive distinct storage accounts and their storage deposits don't collide.
- Introduce `BoundedString<const MAX: usize>` in `omni-types` — a length-capped string wrapper that enforces its bound on construction, JSON deserialization, and Borsh deserialization. Oversized inputs fail with a new `TypesError::StringTooLong`.
- Apply `BoundedString` to `InitTransferMsg`:
  - `msg: Option<BoundedString<MAX_INIT_TRANSFER_MSG_LEN>>` (2048 B) — caps the caller-supplied destination-chain hook payload.
  - `external_id: Option<BoundedString<MAX_EXTERNAL_ID_LEN>>` (64 B) — caps the new identifier.
- Thread `external_id` through `TransferMessage::calculate_storage_account_id` and `TransferMessageStorageAccount::id`: when present, its bytes are appended to the borsh-encoded storage-account struct before hashing.

## Why

Two motivations:

1. **`external_id`**: without it, two identical outbound transfers from the same sender hash to the same virtual storage account — their pre-funded storage deposits collide. `external_id` is a caller-provided discriminator that breaks that collision.
2. **Bounded strings**: `msg` and `external_id` are user-controlled. Without a cap, callers could submit arbitrarily large payloads, inflating storage/gas costs and growing the storage-account-hash input without bound. Enforcing the limit at the type boundary rejects oversized input before any contract logic runs.

## Notes

- `BoundedString` bounds are enforced at every entry point: `BoundedString::new`, `FromStr`, `serde::Deserialize`, `BorshDeserialize`.
- `From<BoundedString<MAX>> for String` + `Default` keep downstream conversion ergonomic (`.map(String::from)`, `.unwrap_or_default()`).
- Chosen limits — 2048 B for `msg`, 64 B for `external_id` — are generous enough for real use (JSON hook payloads, UUIDs / hex-encoded 32-byte hashes) while still bounded.

## Test plan

- [x] New unit tests in `bounded_string.rs` cover: construction within/at/over limit, JSON roundtrip + oversize rejection, Borsh roundtrip + oversize rejection.
- [x] New integration test `test_init_transfer_with_external_id` runs two full init-transfer flows with different `external_id`s and asserts sender/locker/relayer balances reflect both transfers.
- [x] Updated every existing `InitTransferMsg` construction in `omni-tests` and `omni-bridge` tests with `external_id: None`.
- [x] `make clippy-near` clean (pedantic).

### DIFF
--- a/near/omni-tests/src/init_transfer.rs
+++ b/near/omni-tests/src/init_transfer.rs
@@ -17,7 +17,8 @@ mod tests {
         environment::TestEnvBuilder,
         helpers::tests::{
             account_n, build_artifacts, eth_eoa_address, eth_factory_address,
-            get_claim_fee_args_near, get_event_data, relayer_account_id, BuildArtifacts,
+            execution_contains_log, get_claim_fee_args_near, get_event_data, relayer_account_id,
+            BuildArtifacts,
         },
     };
 
@@ -653,7 +654,7 @@ mod tests {
         let status_b = submit_ft_transfer_call(msg_b.clone()).await?;
 
         // Fast-forward a few blocks so both yield/resume callbacks are processed.
-        env.worker.fast_forward(5);
+        env.worker.fast_forward(5).await?;
 
         let required_balance_init_transfer: NearToken = env
             .locker_contract
@@ -679,6 +680,13 @@ mod tests {
         // Each ft_transfer_call should now unblock — resumed by its storage_deposit.
         let result_a = status_a.await?.into_result()?;
         let result_b = status_b.await?.into_result()?;
+
+        for result in [&result_a, &result_b] {
+            assert!(
+                execution_contains_log(result, "Yield init transfer until storage is available at"),
+                "Expected yield log not found in receipt outcomes"
+            );
+        }
 
         // Sign + claim fee for each transfer.
         for result in [&result_a, &result_b] {
--- a/near/omni-tests/src/init_transfer.rs
+++ b/near/omni-tests/src/init_transfer.rs
@@ -8,7 +8,8 @@ mod tests {
     use near_workspaces::{result::ExecutionSuccess, types::NearToken, AccountId};
     use omni_types::{
         near_events::OmniBridgeEvent, BoundedString, BridgeOnTransferMsg, ChainKind, Fee,
-        InitTransferMsg, OmniAddress, TransferId, TransferMessage, UpdateFee,
+        InitTransferMsg, OmniAddress, TransferId, TransferMessage, TransferMessageStorageAccount,
+        UpdateFee,
     };
     use rstest::rstest;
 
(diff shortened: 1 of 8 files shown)


near/omni-types/src/lib.rs
```diff
@@ -566,8 +581,12 @@ pub struct TransferMessageStorageAccount {
 
 impl TransferMessageStorageAccount {
     #[allow(clippy::missing_panics_doc)]
-    pub fn id(&self) -> AccountId {
-        let hash = utils::sha256(&borsh::to_vec(self).unwrap());
+    pub fn id(&self, external_id: Option<String>) -> AccountId {
+        let mut bytes = borsh::to_vec(self).unwrap();
+        if let Some(external_id) = external_id {
+            bytes.extend_from_slice(external_id.as_bytes());
```
frolvanya, line 587: Is it okay that `id(None)` and `id(Some(""))` are identical here, since bytes array will be empty?

near/omni-tests/src/init_transfer.rs
```diff
@@ -545,6 +551,191 @@ mod tests {
...
+                }))
+                .deposit(NearToken::from_yoctonear(1))
+                .max_gas()
+                .transact_async()
+        };
+        let status_a = submit_ft_transfer_call(msg_a.clone()).await?;
+        let status_b = submit_ft_transfer_call(msg_b.clone()).await?;
+
+        // Give the sandbox a few blocks to register both yielded promises before resuming.
+        tokio::time::sleep(std::time::Duration::from_secs(3)).await;
```
frolvanya: I know that it's test and not very critical (it can be flaky), but why do we use sleep instead of fast forward here?

karim-en: ebded5c1a5a82103152b4068510b4677071b4e4d

near/omni-types/src/lib.rs
karim-en: We could replace it with bytes.extend_from_slice(&borsh::to_vec(&external_id).unwrap());, but the downside is that it could break some transfers during the contract upgrade. The probability is low, but it is still possible.

@kiseln what do you think?

kiseln: I'd leave as it is

near/omni-tests/src/init_transfer.rs
```diff
@@ -545,6 +551,191 @@ mod tests {
...
+                }))
+                .deposit(NearToken::from_yoctonear(1))
+                .max_gas()
+                .transact_async()
+        };
+        let status_a = submit_ft_transfer_call(msg_a.clone()).await?;
+        let status_b = submit_ft_transfer_call(msg_b.clone()).await?;
+
+        // Fast-forward a few blocks so both yield/resume callbacks are processed.
+        env.worker.fast_forward(5);
```
kiseln: Do we want to assert that the transfers indeed yielded execution?

kiseln (approved):

near/omni-tests/src/init_transfer.rs
frolvanya: and if not? What to do? Wait more and hope that it'll be yielded a bit later? This test is flaky by design, I don't think that extra checks will help

But so far I see that tests are passing, so maybe we can leave it as is

kiseln: If we accidentally give users more storage the test will pass with a normal flow without yield-resume

karim-en: Fixed in 7fd9c021da913ad84d7b384b48ea99faf18f0e76

kiseln (approved):

PR #616, opened 2026-04-28
feat(1155): add token id to the metadata

karim-en:
This pull request updates the way token metadata is logged in the `OmniBridge` contract. The main change is to improve the metadata emitted for tokens by including both the token address and token ID in the log, making it easier to identify specific tokens.

**Metadata logging improvements:**

* The `name` variable now concatenates the token address and token ID (as a string) separated by a `#`, and this value is used in both the `logMetadataExtension` call and the `LogMetadata` event emission. This replaces the previous approach, which only included the token address.

### DIFF
--- a/evm/src/omni-bridge/contracts/OmniBridge.sol
+++ b/evm/src/omni-bridge/contracts/OmniBridge.sol
@@ -249,19 +249,15 @@ contract OmniBridge is
             }
         }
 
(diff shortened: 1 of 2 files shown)


frolvanya (commented):
Please fix tests

frolvanya (approved):

PR #624, opened 2026-05-28
feat(evm): add coreReceiveWithData callback to HlBridgeToken

karim-en:
This pull request introduces a new callback mechanism to the `HyperliquedBridgeToken` contract, allowing it to receive and process cross-chain messages from HyperCore, with robust authorization, action dispatch, and bridging logic. It also adds comprehensive tests to ensure correct behavior for all new code paths.

**Core contract enhancements:**

* Added the `coreReceiveWithData` function to `HyperliquedBridgeToken`, implementing the `ICoreReceiveWithData` interface. This function dispatches actions based on the first byte of `data`, supporting both direct token transfers and cross-chain bridging via `OmniBridge.initTransfer`, with strict authorization and error handling. [[1]](diffhunk://#diff-e523b53d18de81376b70c2efc3182c67d9bcdf2d84e8aa55db6e4761bdd1407cR4-R50) [[2]](diffhunk://#diff-e523b53d18de81376b70c2efc3182c67d9bcdf2d84e8aa55db6e4761bdd1407cR75-R122)
* Introduced new events (`CoreReceived`) and custom errors (`NotSystemAddress`, `EmptyActionData`, `UnknownAction`) to improve observability and error reporting in the contract.

**Testing improvements:**

* Added a new test suite (`HlBridgeToken.ts`) that covers all major code paths of `HyperliquedBridgeToken`, including authorization checks, action dispatch, error cases, and integration with the real `OmniBridge` contract.
* Tests verify correct minting, ownership restrictions, dispatch logic for both transfer and bridge actions, event emission, and handling of edge cases such as balance insufficiency and type overflows.

### DIFF
--- a/evm/src/omni-bridge/contracts/HlBridgeToken.sol
+++ b/evm/src/omni-bridge/contracts/HlBridgeToken.sol
@@ -1,13 +1,52 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity ^0.8.24;
 
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import {BridgeToken} from "./BridgeToken.sol";
+
+interface IOmniBridgeInitTransfer {
+    function initTransfer(
+        address tokenAddress,
+        uint128 amount,
+        uint128 fee,
+        uint128 nativeFee,
+        string calldata recipient,
+        string calldata message
+    ) external payable;
+}
+
+interface ICoreReceiveWithData {
+    function coreReceiveWithData(
+        address from,
+        bytes32 destinationRecipient,
+        uint32 destinationChainId,
+        uint256 amount,
+        uint64 coreNonce,
+        bytes calldata data
+    ) external;
+}
 
 /// @notice Hyperliquid-specific BridgeToken with two mint paths:
 /// - 2-arg mint(address, uint256): mints on HyperEVM (tokens go directly to user)
 /// - 3-arg mint(address, uint256, bytes): mints on HyperCore (includes _update to system address for spot-balance tracking)
-contract HyperliquedBridgeToken is BridgeToken {
+contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
+    using SafeCast for uint256;
+
     address internal _systemAddress;
+
+    uint8 public constant ACTION_TRANSFER = 0;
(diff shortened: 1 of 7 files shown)


evm/src/omni-bridge/contracts/HlBridgeToken.sol
```diff
@@ -33,4 +72,62 @@ contract HyperliquedBridgeToken is BridgeToken {
...
+    function coreReceiveWithData(
+        address from,
+        bytes32 /*destinationRecipient*/,
+        uint32 /*destinationChainId*/,
+        uint256 amount,
+        uint64 /*coreNonce*/,
+        bytes calldata data
+    ) external override {
+        if (msg.sender != _systemAddress) revert NotSystemAddress();
+        if (data.length == 0) revert EmptyActionData();
```
olga24912, line 106: Should we add a check on the size of data to prevent it from being too large?

evm/src/omni-bridge/contracts/HlBridgeToken.sol
```diff
@@ -1,14 +1,53 @@
...
+    ) external;
+}
+
 /// @notice Hyperliquid-specific BridgeToken with two mint paths:
 /// - 2-arg mint(address, uint256): mints on HyperEVM (tokens go directly to user)
 /// - 3-arg mint(address, uint256, bytes): mints on HyperCore (includes _update to system address for spot-balance tracking)
-contract HyperliquedBridgeToken is BridgeToken {
+contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
+    using SafeCast for uint256;
+
```
olga24912, line 34: Should we add __gap here or in BridgeToken? In case of new fields in BridgeToken?

olga24912 (approved):

evm/src/omni-bridge/contracts/HlBridgeToken.sol
karim-en: No needs, in this case it will be much better to fully copy past the token implementation in case if we will need to add extra fields there

evm/src/omni-bridge/contracts/HlBridgeToken.sol
karim-en: There is no any benefits of doing it, this usually a problem in Near contracts

PR #626, opened 2026-06-05
feat: Aptos support

karim-en:
# feat: Aptos support

Adds the Aptos side of the Omni Bridge — a complete Move package that
mirrors the EVM and Starknet implementations — plus the NEAR-side
`ChainKind::Aptos` wiring so transfers can flow end-to-end.

### Aptos Move package ([aptos/](aptos/))

| Module | Purpose |
|---|---|
| `omni_bridge::omni_bridge` | Main contract — `initialize`, `init_transfer`, `fin_transfer`, `deploy_token`, `log_metadata`, `set_token_metadata`, pause, role management, views |
| `omni_bridge::bridge_token` | Fungible Asset wrapper holding `MintRef`/`BurnRef`/`TransferRef`/`MutateMetadataRef` as a private resource; `package fun` interface |
| `omni_bridge::bridge_types` | Payload structs + Borsh encoders byte-compatible with Starknet |
| `omni_bridge::borsh` | Borsh sequence helpers (`encode_string`, `encode_byte_vec`); primitives delegate to `bcs::to_bytes` |
| `omni_bridge::utils` | Eth-style ECDSA verification via `aptos_std::secp256k1` + keccak256 |

### NEAR-side integration

- `ChainKind::Aptos = 13` and `OmniAddress::Aptos(AptosAddress)` (`H256` alias)
- `OmniAddress::FromStr` accepts `"aptos:0x..."`
- `get_native_token_address(Aptos)` returns the canonical APT FA metadata address (`0xa`)
- Factory address helper for integration tests

### CI

- New [.github/workflows/aptos.yaml](.github/workflows/aptos.yaml) runs `compile` + `test` on PR
- [.github/workflows/update-contracts.yaml](.github/workflows/update-contracts.yaml) extended to build the Aptos package on release tags and include `.mv` modules + `package-metadata.bcs` in the release zip

## Architecture highlights

- **Named-object pattern**: `BridgeState` lives on a deterministic Object owned by `@omni_bridge` (seed `b"omni_bridge::state"`). `disable_ungated_transfer` pins it permanently — deployer-key compromise can't move locked funds.
- **Role-based access control**: `Table<u8, vector<address>>` — three roles (`Admin`, `Pauser`, `MetadataAdmin`) with multi-holder support, generic `grant_role`/`revoke_role`, last-admin guard prevents bricking. Discoverable off-chain via `all_roles()`.
- **Deterministic FA addresses**: raw NEAR token id used as the `create_named_object` seed — no keccak (Aptos's seed is arbitrary-length, unlike Starknet's `felt252`).
- **Capability confinement**: token mint/burn/metadata refs are held in a private resource on each FA object, accessible only via `package fun` from the `omni_bridge` package.

## Borsh compatibility

The Borsh payload encoding is byte-identical to the Starknet variant so the NEAR MPC signature works without per-chain branching on the NEAR side:
- 32-byte big-endian addresses
- Destination chain id interleaved as the OmniAddress tag before `token_address` and `recipient`
- Destination chain id is bound into the hash (not the payload) → prevents cross-chain replay
- `message` field deliberately untagged (matches Starknet)

## Security invariants

- **No replay**: `destination_nonce` checked + marked before any token transfer; `origin_nonce` incremented atomically
- **State before external calls**: pause/nonce/role checks all happen before mint or transfer
- **No token release without signature**: every mint/transfer path goes through eth-signature verification
- **Event completeness**: `InitTransfer` carries every field the NEAR side needs to reconstruct the transfer

## Roles & operational model

| Role | Authority |
|---|---|
| `Admin` | Rotate any role, change `near_bridge_derived_address`, set pause flags |
| `Pauser` | `pause_all` only |
| `MetadataAdmin` | `set_token_metadata` (icon URI / project URI) |

Each role accepts an unbounded list of holders. Suggested ops setup: cold-multisig Admin, warm-multisig Pauser, smaller-key MetadataAdmin.

## Testing

35 unit tests covering:

- Borsh encoding edge cases (length-prefix exhaustiveness, empty inputs, multi-byte lengths)
- Payload Borsh layout (offsets, dual chain-id tag, Option semantics)
- Nonce bitmap (word-boundary at 127/128, idempotent mark)
- Pause flags + access control
- Role grant/revoke (idempotent grant, last-admin guard, admin step-down)
- Metadata mutation (admin only, bridge-tokens only)
- Bridge token mint/burn round-trip
- Signature verification negatives (invalid sig, wrong length)
- **`init_transfer`** (7 tests): bridged-token burn, non-bridge-token lock, nonce increment, paused, zero amount, fee≥amount, amount>u64::MAX

Run with:
```sh
aptos move test --named-addresses omni_bridge=0xCAFE

### CODE REVIEW

**Comment by karim-en** [comment]
@copilot resolve the merge conflicts in this pull request

File: aptos/sources/utils.move [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+    use aptos_std::secp256k1;
+
+    /// Signature payload could not be parsed.
+    const E_INVALID_SIGNATURE_LENGTH: u64 = 1;
+    /// `ecdsa_recover` failed to recover a public key.
+    const E_RECOVER_FAILED: u64 = 2;
+    /// Recovered Ethereum address does not match the expected signer.
+    const E_INVALID_SIGNATURE: u64 = 3;
+
+    const MAX_ALLOWED_DECIMALS: u8 = 18;
```
kiseln: With u64 to store balances this is incompatible. I suggest we use the standard 8 decimals

aptos/sources/omni_bridge.move
```diff
@@ -0,0 +1,792 @@
...
+        /// 20-byte recovered address of the NEAR MPC-derived Ethereum signer.
+        near_bridge_derived_address: vector<u8>,
+        /// Chain id of this bridge instance (mixed into transfer payload hashes
+        /// to prevent cross-chain replay).
+        chain_id: u8,
+        /// Monotonically increasing origin nonce assigned to outbound transfers.
+        current_origin_nonce: u64,
+        /// Bitmap of finalized destination nonces. `slot = nonce / 128`,
+        /// `bit = nonce % 128`. Persists in a `Table<u64, u128>`.
+        completed_transfers: Table<u64, u128>,
```
kiseln, line 109: Could you remind me why did we use bitmap for nonces on Starknet? Is it because of gas cost? And does the same issue apply on Aptos?

kiseln (changes requested):

aptos/sources/utils.move
karim-en: 8c2886043e366bafa810ca3e757f301b8c0df4fc

aptos/sources/omni_bridge.move
karim-en: It is for storage/gas cost optimization

kiseln (approved):

envin3 (approved):

frolvanya (approved):

PR #629, opened 2026-06-15
feat: added aptos in mpc prover and bumped rust version to 1.96.0 and near-plugins to 0.5.2

frolvanya:
Nearcore's migration to wasmtime in 2.12.0 unpinned rust version that can be used for smart contract development. Repos like mpc already migrated to near-sdk-rs v5.28.0+, which bumped msrv from 1.86.0 to 1.88.0, so in order to keep compatibility we need to bump rust version in our contracts as well

### DIFF
--- a/near/token-deployer/src/lib.rs
+++ b/near/token-deployer/src/lib.rs
@@ -26,7 +26,10 @@ pub enum Role {
 
 #[near(contract_state)]
 #[derive(Pausable, Upgradable, PanicOnDefault)]
-#[pausable(pause_roles(Role::PauseManager), unpause_roles(Role::UnpauseManager))]
+#[pausable(
+    pause_roles(Role::PauseManager),
+    unpause_roles(Role::DAO, Role::UnpauseManager)
+)]
 #[access_control(role_type(Role))]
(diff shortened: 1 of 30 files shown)


near/token-deployer/src/lib.rs
```diff
@@ -21,11 +21,12 @@ pub enum Role {
     UpgradableCodeDeployer = 4,
     Controller = 5,
     LegacyController = 10,
+    UnpauseManager = 11,
 }
 
 #[near(contract_state)]
 #[derive(Pausable, Upgradable, PanicOnDefault)]
-#[pausable(manager_roles(Role::PauseManager))]
+#[pausable(pause_roles(Role::PauseManager), unpause_roles(Role::UnpauseManager))]
```
karim-en: we should add DAO to unpause_roles, the UnpauseManager isn't really needed, but we can keep it

karim-en (changes requested):

near/token-deployer/src/lib.rs
frolvanya: Added in dce7eb295aa31c419941b6df09be8f970845218b

PR #633, opened 2026-07-06
feat: added sui support

frolvanya:
Adds the Sui side of the Omni Bridge: a Move package mirroring the
Aptos/Starknet contracts, plus the NEAR-side type wiring and CI.

## What's inside

- `sui/` Move package:
  - `init_transfer` — send tokens Sui → NEAR: locks native coins in bridge
    custody (or burns bridge-deployed ones), collects an optional SUI
    native fee, and emits an `InitTransfer` event for the NEAR MPC to read
    (no Wormhole).
  - `fin_transfer` — receive tokens NEAR → Sui: verifies the NEAR MPC
    secp256k1 signature over the borsh payload, replay-protected by a
    destination-nonce bitmap, then unlocks or mints to the recipient.
  - `deploy_token` / `log_metadata` — token registration in both
    directions. Sui can't create coins at runtime (one-time-witness rule),
    so NEAR-originated tokens are published from `sui/token_template/` and
    bound to the MPC-signed metadata payload (zero-supply TreasuryCap +
    surrendered UpgradeCap + metadata equality checks).
  - Role-based admin (Admin/Pauser/MetadataAdmin), pause flags, and a
    version-gated shared state with `migrate` for Sui's package-upgrade
    model.
- Sui-specific wire format: coins are types, not addresses, so the 32-byte
  token id is `keccak256(canonical coin type string)`; payload layout is
  byte-identical to Aptos. Chain id = 14 (`ChainKind::Sui`), decimals
  clamped to 9.
- NEAR-side wiring in `omni-types`/`omni-bridge`: `ChainKind::Sui`,
  `OmniAddress::Sui(H256)`, token prefix, native-token constant, tests.
  (Event parsers / mpc-prover dispatch are follow-ups, blocked on near/mpc
  adding Sui read support.)
- CI: `sui.yaml` (build + test) and Aptos/Sui release artifacts in
  `update-contracts.yaml`.

## Testing

- 86 Move unit tests, including byte-exact borsh layout checks and real
  secp256k1 signature vectors (positive and negative), generated offline
  against an independent encoder implementation.
- Full e2e smoke test on a local Sui network: publish → initialize →
  log_metadata → init_transfer (lock) → signed fin_transfer (unlock) →
  replay correctly rejected → template publish → deploy_token →
  fin_transfer mint → init_transfer burn. Lock/unlock and mint/burn paths
  both verified on-chain, and event `token_address` values matched the
  offline keccak ids exactly.
- A test version is deployed on Sui testnet (initialized with a test MPC
  key — not for production use):
  - package: 0x587d69879cb01f62a88526a7448424c37cf767387473c1e610bd007dc95fa6ba
  - BridgeState: 0xcbe9d3824224eaa23e2ef70667fff857ddd0929b7c0e42935d509347505fe858
  - example init_transfer (0.001 SUI → near:frolik.testnet):
    https://testnet.suivision.xyz/txblock/HcrYCissQ4qS1bF5DXa5zmejYiHVwupPWMuq1t9eYiot

See `sui/README.md` for the trust model, the documented deploy_token
front-running trade-off, and the deployment guide in `sui/deployment.md`.


near/omni-types/src/lib.rs
```diff
@@ -183,6 +189,10 @@ pub type EvmAddress = H160;
 pub type UTXOChainAddress = String;
 pub type StarknetAddress = H256;
 pub type AptosAddress = H256;
+/// Sui coins are types, not addresses: for tokens this carries
+/// `keccak256(canonical coin type string)`; for accounts, the native
+/// 32-byte Sui address.
+pub type SuiAddress = H256;
```
frolvanya, line 195: Sui treats coins differently as in Aptos, I decided to keep H256 for more similarity between move-related chains and store hash as an address, but maybe it's more transparent to use bounded string here

sui/sources/utils.move
```diff
@@ -0,0 +1,102 @@
...
+/// `secp256k1_ecrecover` hashes the message internally (flag 0 =
+/// keccak256), so `message_bytes` is the RAW borsh payload — never a
+/// digest. The Ethereum address is the last 20 bytes of
+/// `keccak256(uncompressed_pubkey[1..65])`.
+public fun verify_eth_signature(
+    message_bytes: &vector<u8>,
+    signature: &vector<u8>,
+    expected_address: &vector<u8>,
+) {
+    assert!(signature.length() == 65, E_INVALID_SIGNATURE_LENGTH);
```
frolvanya, line 47: FYI, signature of this method is a bit different from Aptos, since we pass `v` as a last byte

frolvanya (commented):
A bit more information regarding deployment, testing and multisig:

# Installation

https://docs.sui.io/getting-started/onboarding/sui-install#quick-install

```bash
$ curl -sSfL https://raw.githubusercontent.com/Mystenlabs/suiup/main/install.sh | sh
$ suiup install sui@testnet # or mainnet
$ suiup default set sui@testnet # or mainnet
```

Some usefull initial commands:
```bash
$ sui client active-address
$ sui client faucet # or https://faucet.suilearn.io/
$ sui client balance
```

# Package deployment

1) First we need to publish package:
```bash
$ sui client publish
```
2) Then call `initialize`: 
```bash
sui client call --package <Package address> --module omni_bridge --function initialize --args <BridgeState> <NEAR MPC derived address (20 bytes)> 14
```
> [!NOTE]
> Both package address and bridge state could be found after executing first command (`sui client publish`)
> The `14` is the `ChainKind::Sui` discriminant; a wrong value can be corrected later with `set_chain_id` (admin).

# Calling methods

```bash
$ sui client ptb --split-coins gas "[1000000]" --assign pay --move-call 0x2::coin::zero "<0x2::sui::SUI>" --assign zero_fee --move-call <Package address>::omni_bridge::init_transfer "<0x2::sui::SUI>" @<BridgeState> pay.0 0 zero_fee '"near:frolik.testnet"' "vector[]"

# https://testnet.suivision.xyz/txblock/HcrYCissQ4qS1bF5DXa5zmejYiHVwupPWMuq1t9eYiot?tab=Overview
```

# Multisig 

https://docs.sui.io/develop/transactions/transaction-auth/multisig

sui/sources/utils.move
```diff
@@ -0,0 +1,102 @@
...
+    assert!(expected_address.length() == 20, E_INVALID_SIGNATURE);
+
+    let mut sig = *signature;
+    let v = &mut sig[64];
+    if (*v >= 27) {
+        *v = *v - 27;
+    };
+
+    let compressed = ecdsa_k1::secp256k1_ecrecover(&sig, message_bytes, 0);
```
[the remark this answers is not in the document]
frolvanya: Technically Aptos has the same "flaw": https://github.com/Near-One/omni-bridge/blob/f499d46c2d58869e4cbb5d19cff703ff7d2731d6/aptos/sources/utils.move#L47-L48
and starknet has the loosest check: https://github.com/Near-One/omni-bridge/blob/f499d46c2d58869e4cbb5d19cff703ff7d2731d6/starknet/src/omni_bridge.cairo#L398-L406

But I'm not sure how critical is that
cc @karim-en

frolvanya:
@claude review

frolvanya:
@claude review

sui/sources/borsh.move
```diff
@@ -0,0 +1,32 @@
...
+    let len = val.length() as u32;
+    let mut result = vector[
+        (len & 0xFF) as u8,
+        ((len >> 8) & 0xFF) as u8,
+        ((len >> 16) & 0xFF) as u8,
+        ((len >> 24) & 0xFF) as u8,
+    ];
+    result.append(*val);
+    result
+}
```
karim-en:
```suggestion
use std::bcs;
use std::string::String;

/// Borsh-style byte vector: 4-byte little-endian length + bytes.
public fun encode_byte_vec(val: &vector<u8>): vector<u8> {
    let mut result = bcs::to_bytes(&(val.length() as u32));
    result.append(*val);
    result
}

/// Borsh-style string: 4-byte little-endian length + UTF-8 bytes.
public fun encode_string(val: &String): vector<u8> {
    encode_byte_vec(val.as_bytes())
}

```

sui/sources/omni_bridge.move
```diff
@@ -0,0 +1,1128 @@
...
+
+/// Top-level bridge state: a single shared object created at publish.
+public struct BridgeState has key {
+    id: UID,
+    /// See `VERSION`.
+    version: u64,
+    /// Role discriminant -> list of holder addresses. Each role can
+    /// have any number of holders; all of them are equally privileged.
+    /// See the `ROLE_*` constants for the discriminant values.
+    roles: Table<u8, vector<address>>,
```
karim-en, line 151: Is there any advantage of using custom roles over libs https://docs.openzeppelin.com/contracts-sui?

sui/sources/omni_bridge.move
```diff
@@ -0,0 +1,1128 @@
...
+
+/// Recover the caps from a `TokenSetup` that can never be bound - metadata
+/// that will never match the signed payload, or a `VERSION` bump before the
+/// bind (which strands pre-published deploy inventory). `TokenSetup` has no
+/// `drop`, so without this the caps are lost.
+///
+/// Cancelling retires the coin type: a `TokenSetup` only comes from
+/// `prepare_token` and the witness is already spent, so the released caps
+/// govern a currency the bridge can never adopt.
+public fun cancel_token_setup<T>(setup: TokenSetup<T>): (TreasuryCap<T>, MetadataCap<T>) {
```
karim-en, line 522: Who can cancel it and when?

sui/sources/utils.move
```diff
@@ -0,0 +1,105 @@
...
+/// Verify an Ethereum-style 65-byte signature (`r || s || v`) over
+/// `message_bytes`.
+///
+/// `v` is the Ethereum recovery id as emitted by the NEAR MPC
+/// (`recovery_id + 27`); 0/1 are also accepted. Sui's
+/// `secp256k1_ecrecover` hashes the message internally (flag 0 =
+/// keccak256), so `message_bytes` is the RAW borsh payload — never a
+/// digest. The Ethereum address is the last 20 bytes of
+/// `keccak256(uncompressed_pubkey[1..65])`.
+public fun verify_eth_signature(
```
karim-en, line 42: Two things to mentation:
- it it is ok to use the full ecdsa validation instead of Ethereum
- even with Ethereum version it is possible to improve the code readability:

```
public fun verify_eth_signature(
    message_bytes: &vector<u8>,
    signature: &vector<u8>,
    expected_address: &vector<u8>,
) {
    assert!(signature.length() == 65, E_INVALID_SIGNATURE_LENGTH);
    assert!(expected_address.length() == 20, E_INVALID_SIGNATURE);

    let mut sig = *signature;
    let v = sig.pop_back();
    sig.push_back(if (v >= 27) v - 27 else v);

    let uncompressed = ecdsa_k1::decompress_pubkey(
        &ecdsa_k1::secp256k1_ecrecover(&sig, message_bytes, 0),
    );

    // `skip(1)` drops the 0x04 uncompressed-point prefix, leaving the raw
    // 64-byte key Ethereum hashes; `skip(12)` keeps the low 20 bytes of the
    // digest, which is the address.
    let digest = hash::keccak256(&uncompressed.skip(1));
    assert!(digest.skip(12) == *expected_address, E_INVALID_SIGNATURE);
}
```

karim-en (approved):

karim-en (changes requested):
Approved by mistake

sui/sources/omni_bridge.move
frolvanya: It's not under maintenance at all: https://github.com/OpenZeppelin/contracts-sui#openzeppelin-contracts-for-sui

Also, as you mentioned I couldn't find `pausable` as well, so we would use it only for access control. IMO not worth it

sui/sources/omni_bridge.move
frolvanya: Whoever has a possession of `TokenSetup`

sui/sources/utils.move
frolvanya: Applied refactor in [002b0ef](https://github.com/Near-One/omni-bridge/pull/633/commits/002b0efcd523bdc4bca77d886c272f0a7a1940a0)

sui/sources/omni_bridge.move
frolvanya: Added comment in code in [d8927c0](https://github.com/Near-One/omni-bridge/pull/633/commits/d8927c08427eac661a355e4e51fb9b3384b9e9a2)

frolvanya:
@claude review

karim-en:
@frolvanya can you please setup fmt checks and tests to CI for SUI

frolvanya:
> @frolvanya can you please setup fmt checks and tests to CI for SUI

@karim-en we already had build+test CI for SUI, but added fmt in [db00aee](https://github.com/Near-One/omni-bridge/pull/633/commits/db00aeecb21c049123be260d551622d7591e916a)

karim-en (approved):

sui/sources/omni_bridge.move
```diff
@@ -0,0 +1,1112 @@
...
+
+/// Permissionless: emit a `LogMetadata` event describing an existing coin.
+/// The NEAR side picks this event up (via MPC read) to decide whether to
+/// sign a `deploy_token` payload for the mirror token on its side. Also
+/// registers the coin type in `token_registry` so off-chain actors can
+/// resolve the 32-byte token id back to the concrete type.
+public fun log_metadata<T>(state: &mut BridgeState, coin_metadata: &CoinMetadata<T>) {
+    assert_version(state);
+    assert_configured(state);
+
```
commit4ikus, line 404: log_metadata / log_metadata_registry should reject bridge-deployed coins before emitting LogMetadata
evm, aptos, ... have !is_bridge_token

assert!(!is_bridge_token<T>(state), E_TOKEN_EXIST);

sui/sources/omni_bridge.move
```diff
@@ -0,0 +1,1112 @@
...
+// -------- Admin --------
+
+/// Add `new_holder` to the set of `role` holders. No-op if the
+/// address already holds the role. Caller must hold `ROLE_ADMIN`.
+/// Works for every role, including `ROLE_ADMIN` itself.
+public fun grant_role(state: &mut BridgeState, role: u8, new_holder: address, ctx: &TxContext) {
+    assert_version(state);
+    assert_role(state, ROLE_ADMIN, ctx.sender());
+    add_role_holder(state, role, new_holder);
+}
```
commit4ikus, line 320: grant_role / revoke_role update state but emit nothing. All other chains have events. 

Should we keep this consistent across chains?

event::emit(RoleGranted ...

sui/sources/omni_bridge.move
```diff
@@ -0,0 +1,1112 @@
...
+    state.near_bridge_derived_address = near_bridge_derived_address;
+    state.chain_id = chain_id;
+}
+
+// -------- Admin --------
+
+/// Add `new_holder` to the set of `role` holders. No-op if the
+/// address already holds the role. Caller must hold `ROLE_ADMIN`.
+/// Works for every role, including `ROLE_ADMIN` itself.
+public fun grant_role(state: &mut BridgeState, role: u8, new_holder: address, ctx: &TxContext) {
```
commit4ikus, line 314: grant_role / revoke_role accept any role: u8
Lets validate {Admin, Pauser, MetadataAdmin} before updating state

```
fun assert_known_role(role: u8) {
    assert!(
        role == ROLE_ADMIN ||
        role == ROLE_PAUSER ||
        role == ROLE_METADATA_ADMIN,
        E_INVALID_ROLE,
    );
}
```

.github/workflows/update-contracts.yaml
```diff
@@ -67,6 +72,48 @@ jobs:
...
+        run: aptos move compile --save-metadata --named-addresses omni_bridge=0xCAFE
+        timeout-minutes: 30
+
+      - name: Install Sui CLI
+        run: |
+          ASSET="sui-mainnet-v${SUI_VERSION}-ubuntu-x86_64.tgz"
+          URL="https://github.com/MystenLabs/sui/releases/download/mainnet-v${SUI_VERSION}/${ASSET}"
+          curl -fsSL "$URL" -o sui.tgz
+          mkdir -p sui-bin
+          tar -xzf sui.tgz -C sui-bin
```
commit4ikus: Should be `tar -xzf sui.tgz -C sui-bin ./sui` 
*avoids pulling multi-GB debug binaries into the release job

sui/sources/omni_bridge.move
frolvanya: Eh, events are emitted, because we use libraries like OpenZeppelin and near-plugins, but we don't really track/index those events. So I wouldn't add them

cc @karim-en 

P.S: Actually, aptos implements access control manually as well and it has event, so I'm wrong. I'll add event then

.github/workflows/update-contracts.yaml
frolvanya: Forgot to update this workflow after fixing it in `sui.yaml`. Fixed in https://github.com/Near-One/omni-bridge/pull/633/commits/3a022d03061a368eb009b3dc788f38c9465ee69c

sui/sources/omni_bridge.move
frolvanya: Added events in https://github.com/Near-One/omni-bridge/pull/633/commits/dcbedafa55e1802d199de9e6e38ef0cb0a24887d

sui/sources/omni_bridge.move
frolvanya: Added assertion in https://github.com/Near-One/omni-bridge/pull/633/commits/31006cba14805bea4f5749463c301d6a0e68322f

sui/sources/omni_bridge.move
frolvanya: We had similar reports on many other chains. This is by design, methods that are callable by DAO are already dangerous, so it's up to DAO members to verify everything carefully

Moreover, in this case adding a role that doesn't exist is no-op and can be revoked immediately. I think it's fine to skip such checks

frolvanya:
@claude review

commit4ikus (approved):

PR #639, opened 2026-07-30
chore: remove unused features

karim-en:
This pull request introduces several important changes to the `near/omni-bridge` contract, focusing on removing support for the `FastFinTransfer` feature and deprecating the `update_transfer_fee` function. The changes improve contract safety by disabling unsupported or deprecated functionality.

### Removal and Deprecation of Features

* The `update_transfer_fee` method was removed, disabling the ability to update transfer fees after a transfer is initiated. This prevents users from modifying fees, which may be necessary for security or protocol reasons.
* The `FastFinTransfer` variant in `BridgeOnTransferMsg` is no longer supported: any attempt to use it will now cause the contract to panic with an explicit error message, instead of calling the removed `fast_fin_transfer` method. [[1]](diffhunk://#diff-b60c5b1d9739aba26a808fe5bb90473c2c30a608886757604357c66a764256b6L269-L271) [[2]](diffhunk://#diff-b60c5b1d9739aba26a808fe5bb90473c2c30a608886757604357c66a764256b6R281-R283)

### Maintenance

* The workspace version in `near/Cargo.toml` was bumped from `0.4.12` to `0.4.13` to reflect these changes.


near/omni-bridge/src/lib.rs
```diff
@@ -281,6 +278,9 @@ impl Contract {
                     .detach();
                 PromiseOrPromiseIndexOrValue::Value(U128(0))
             }
+            BridgeOnTransferMsg::FastFinTransfer(_) => {
```
frolvanya: Do we even need this and other places that are kept to preserve backward compatibility? No one used this, I think it's better to remove this code entirely instead of leaving dead code everywhere

kiseln: I suppose we can release this simple change faster. We can consider removing outstanding code as a follow up

near/omni-bridge/src/lib.rs
```diff
@@ -749,7 +697,7 @@ impl Contract {
         }
     }
 
-    #[allow(clippy::needless_pass_by_value)]
+    #[allow(clippy::needless_pass_by_value, dead_code)]
     fn fast_fin_transfer(
```
olga24912, line 701: Can we just remove this functions?

olga24912 (approved):

near/omni-tests/src/fast_transfer.rs
```diff
@@ -13,7 +13,7 @@ mod tests {
         locker_args::{FinTransferArgs, StorageDepositAction},
         prover_result::{InitTransferMessage, ProverResult},
         BridgeOnTransferMsg, ChainKind, FastFinTransferMsg, Fee, OmniAddress, TransferId,
-        TransferIdKind, TransferMessage, UnifiedTransferId, UpdateFee,
```
olga24912, line 16: Maybe we should remove everything related to FastTransfer.

near/omni-bridge/src/lib.rs
frolvanya: It's better to add pause flags for fast methods in this case

near/omni-bridge/src/lib.rs
karim-en: Revereted some changes and added panic da57ef4e40055ffab65bcf81c552f3e673d9a605

near/omni-tests/src/fast_transfer.rs
karim-en: Revereted some changes and added panic da57ef4e40055ffab65bcf81c552f3e673d9a605

near/omni-bridge/src/lib.rs
karim-en: Revereted some changes and added panic da57ef4e40055ffab65bcf81c552f3e673d9a605

karim-en:
@claude review

PR #647, opened 2026-08-21
feat(near): added sui support in mpc prover

frolvanya:
## Summary

Adds the NEAR-side prover path for Sui, so `InitTransfer` / `FinTransfer` /
`DeployToken` / `LogMetadata` events emitted by the Sui locker can be proven
back to NEAR. This was the last code blocker listed in `sui/README.md`;
everything remaining after this is DAO calls and deployment.

Mirrors the Aptos rollout.

## What's included

- **`near/omni-types/src/sui/events.rs`** — parsers for the four bridge
  events, plus `parse_sui_event` / `parse_sui_proof` dispatch.
- **`near/omni-types/src/sui/bcs.rs`** — minimal BCS reader (no new deps).
- **`MpcFinality::Sui(SuiFinality::Checkpointed)`** and the `ChainKind::Sui`
  arm in `mpc-omni-prover` (`request_to_chain_kind`,
  `request_matches_finality`, `verify_callback`, seeded in `init`).
- **`near-mpc-sdk` rev bump** `dcc9e042` -> `70c40b0f` (merge commit of
  near/mpc#3754, which adds the Sui contract-interface DTOs). The old pin has
  no Sui types at all. Verified no breaking changes for EVM / Starknet /
  Aptos.

## Two things worth reviewing closely

**1. BCS, not JSON.** Aptos delivers `AptosEvent.data` as a JSON rendering;
Sui delivers `SuiEvent.bcs` as canonical BCS bytes. Parsing is therefore
positional and must match the Move struct declaration field-for-field — note
the Sui events carry an extra `coin_type: String` between `token_address` and
the numeric fields, which Aptos does not have.

Two guards make a layout mistake loud instead of silent:
  - `Reader::finish()` rejects trailing bytes, so a wrong field order fails
    rather than producing a plausible-looking wrong value;
  - ULEB128 lengths are canonical-only (BCS uses ULEB128 where Borsh uses a
    fixed u32 — easy to get wrong).

**2. Emitter comes from `type_tag`, not `event.package_id`.** A Move type
keeps its *defining* package id across package upgrades, while `package_id`
becomes the upgraded package's id. Only the former stays equal to the factory
address registered on NEAR, so this survives a bridge package upgrade. Same
rule as the Aptos type-tag-address convention.

## One extra check Aptos can't do

Sui coins are types, so every event carries both the 32-byte wire id and the
`coin_type` string it derives from. The parsers assert
`keccak256(coin_type) == token_address`, binding the two representations.

## Tests

113 passing (+25): 77 in `omni-types`, 36 in `mpc-omni-prover`.

BCS fixtures are generated by an independent Python reference encoder, so the
tests don't just assert the Rust against itself. Coverage: all four events,
tag dispatch, proof-kind/tag mismatch, the keccak binding (flip one byte of
the token id), trailing and truncated payloads, short-form addresses,
malformed tags, non-canonical ULEB128, non-UTF-8 strings, and the
prover-level dispatch/finality/roundtrip cases mirroring Aptos.

Verified: `cargo test`, `make clippy-near` (pedantic, `-D warnings`),
`make fmt-near`, and `cargo check --target wasm32-unknown-unknown`.

### DIFF
--- a/near/CLAUDE.md
+++ b/near/CLAUDE.md
@@ -60,11 +60,11 @@ Proxy to Wormhole protocol for chains without light clients (Solana, BNB, EVM L2
 4. Return typed `ProverResult`
 
 ### mpc-omni-prover
-Verifies foreign chain events by calling the NEAR MPC network's `verify_foreign_transaction` API on-chain. The prover initiates the MPC verification as a cross-contract call and validates the response in a callback. Each deployed instance is configured for a specific chain and finality level via `MpcFinality`. Supports any chain supported by the MPC network (currently EVM chains and Starknet, extensible to others).
+Verifies foreign chain events by calling the NEAR MPC network's `verify_foreign_transaction` API on-chain. The prover initiates the MPC verification as a cross-contract call and validates the response in a callback. Each deployed instance is configured for a specific chain and finality level via `MpcFinality`. Supports any chain supported by the MPC network (currently EVM chains, Starknet, Aptos and Sui).
 
 **State:**
 - `mpc_contract_id` — AccountId of the MPC signer contract (e.g. `v1.signer`)
-- `finality` — `MpcFinality::Evm(EvmFinality)` or `MpcFinality::Starknet(StarknetFinality)`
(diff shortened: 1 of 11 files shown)


frolvanya:
@claude review

PR #650, opened 2026-09-09
poc: two step init-transfer

karim-en:


evm/src/omni-bridge/contracts/OmniBridgeWormholeDeferred.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {OmniBridgeWormhole} from "./OmniBridgeWormhole.sol";
+
+// slither-disable-start unused-return
+contract OmniBridgeWormholeDeferred is OmniBridgeWormhole {
```
karim-en, line 7: Add gap to OmniBridgeWormhole or extend the OmniBridgeWormhole with flag instead of inheritance to simplify future migrations.

evm/src/omni-bridge/contracts/OmniBridgeWormholeDeferred.sol
```diff
@@ -0,0 +1,130 @@
...
+                amount,
+                fee,
+                nativeFee,
+                recipient,
+                message
+            );
+    }
+
+    /// @notice Whether `originNonce` still has an unpublished message queued.
+    function isQueued(uint64 originNonce) external view returns (bool) {
```
karim-en: maybe useless

PR #651, opened 2026-09-10
HyperCore->Near two step v2

olga24912:
# Fix stuck HyperCore → NEAR transfers on HyperEVM

## The problem

When a HyperCore user sends tokens to HyperEVM, Hyperliquid runs the callback as a **system transaction**. Those transactions are strange: they aren't listed in their own block's transaction list, and their logs are missing from the block's log index (`eth_getLogs` / `logsBloom`).

Wormhole guardians only ever read that log index. So when the bridge published a message from inside that callback, **no guardian ever saw it** — no signature, no VAA, and the transfer never completed on NEAR. This is permanent, not a delay: re-polling never finds it.

Confirmed on mainnet — tx `0x60f69a38…` in block 45,282,290 burned the tokens and produced a message that was never signed.

## The fix: split the transfer into two steps

**Step 1 — `preInitTransfer`** runs inside the system-transaction callback. It only *records* the transfer and emits `PreInitTransfer`. Nothing is burned, no Wormhole message is sent.

**Step 2 — `triggerPendingInitTransfer`** is called later by anyone, from an ordinary transaction. It checks the details match what was recorded, burns the tokens, publishes the Wormhole message, and emits `InitTransfer`.

Since step 2 is a normal transaction, its logs land in the index and the guardians pick it up.

Regular `initTransfer` — a user calling the bridge directly on HyperEVM — is completely unchanged. Only the HyperCore path is split.

## Design choices worth knowing

- **Step 1 rejects almost nothing.** If it reverted, the tokens would be stranded on HyperCore with no way back. So the pause check and the fee check both live in step 2, where a revert is safe and retryable.
- **Only a hash is stored, not the full transfer.** Storing the whole payload costs ~133k more gas and blows the ~200k gas budget that HyperCore system transactions run under. The hash fits with ~4.8k to spare.
- **The hash uses `abi.encode`, not `encodePacked`** — packed encoding would let `"ab"+"c"` and `"a"+"bc"` collide into the same commitment.
- **A keeper can't find pending work from logs** (same invisibility problem, one layer up). It has to read state: `currentOriginNonce` for the upper bound, then `pendingInitTransfers(nonce)` — non-zero means work to do.
- **Storage is ERC-7201 namespaced**, so it can never collide with variables added to the base contracts later.

## Also fixed

- `HlBridgeToken` now rejects `address(0)` as a transfer recipient. `_update` treats a zero recipient as a burn, which would have silently destroyed tokens that HyperCore still counts as backed.
- `upgrade-bridge-token` was passing the NEAR token id straight into `upgradeToken`, which expects the token's proxy address — it now looks the address up and errors clearly if the token isn't registered. It also takes `--contract` instead of being hardcoded to `BridgeTokenV2`.
- `upgrade-factory` takes `--contract` so it can target `HlOmniBridge`.

## Tests

~15 new cases in `evm/tests/HlBridgeToken.ts`, covering: nothing is published inline, the second step burns and publishes, double-submission is rejected, a tampered payload fails the hash check, an unregistered token can't commit, nonces stay distinct, HyperCore deposits still work while init-transfer is paused, and a bad fee is accepted at step 1 but rejected at step 2.

## Not in this PR

`HlOmniBridge` still extends `OmniBridgeWormhole` and still publishes Wormhole messages. Removing the Wormhole dependency is a separate change that needs the NEAR side to register a different prover for HyperEVM first.

### DIFF
--- a/evm/src/omni-bridge/contracts/HlBridgeToken.sol
+++ b/evm/src/omni-bridge/contracts/HlBridgeToken.sol
@@ -4,15 +4,15 @@ pragma solidity ^0.8.24;
 import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import {BridgeToken} from "./BridgeToken.sol";
 
-interface IOmniBridgeInitTransfer {
-    function initTransfer(
-        address tokenAddress,
+interface IOmniBridgeQueueInitTransfer {
+    function queueInitTransfer(
+        address sender,
+        uint64 coreNonce,
         uint128 amount,
         uint128 fee,
-        uint128 nativeFee,
         string calldata recipient,
         string calldata message
-    ) external payable;
+    ) external returns (uint64 originNonce);
 }
 
 interface ICoreReceiveWithData {
@@ -39,18 +39,6 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
     uint8 public constant ACTION_TRANSFER = 0;
     uint8 public constant ACTION_INIT_TRANSFER = 1;
 
-    /// @notice Id of the next transfer to be committed. Also the relayer's cursor:
-    /// commitments are enumerable by reading this and walking the ids it has not
-    /// submitted yet, which is the only discovery path that does not depend on the
-    /// callback's logs (see `coreReceiveWithData`).
-    uint256 public nextPendingInitTransferId;
-
-    /// @notice id => keccak256(abi.encode(sender, coreNonce, amount, fee, recipient,
-    /// message)) of a transfer awaiting submission; zero means none. Keyed by our own
-    /// counter rather than by `coreNonce`, which HyperCore sequences per sender and
-    /// so does not identify a transfer on its own.
-    mapping(uint256 => bytes32) public pendingInitTransfers;
-
     event CoreReceived(
         address indexed sender,
         uint8 indexed action,
@@ -59,23 +47,9 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
         bytes data
     );
 
-    /// @notice Carries the full record behind a commitment — the canonical source
-    /// for the values a submitter must hand back to `triggerPendingInitTransfer`.
-    event PreInitTransfer(
-        uint256 indexed id,
-        address indexed sender,
-        uint64 indexed coreNonce,
-        uint128 amount,
-        uint128 fee,
-        string recipient,
-        string message
-    );
-
     error NotSystemAddress();
     error EmptyActionData();
     error UnknownAction(uint8 action);
-    error PendingInitTransferNotFound(uint256 id);
-    error PayloadMismatch(uint256 id);
     error InvalidRecipient();
 
     function initialize(
@@ -166,10 +140,9 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
     }
 
     /// @dev Decodes eagerly so a malformed payload reverts while the transfer is
-    /// still atomic with the HyperCore debit. Storing only the commitment keeps this
-    /// callback's storage cost at one slot regardless of payload length, so an
-    /// oversized `message` cannot push it past the HyperEVM small-block gas limit —
-    /// a revert there would strand the tokens on HyperCore with no way to retry.
+    /// still atomic with the HyperCore debit. The commitment itself lives on the
+    /// bridge, keyed by `originNonce`: one contract to monitor instead of one per
+    /// token, and this contract keeps no storage of its own for the flow.
     function _queueInitTransfer(
         address from,
         uint64 coreNonce,
@@ -179,8 +152,7 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
         (uint128 fee, string memory recipient, string memory message) = abi
             .decode(tail, (uint128, string, string));
 
-        uint256 id = nextPendingInitTransferId++;
-        pendingInitTransfers[id] = _initTransferCommitment(
+        IOmniBridgeQueueInitTransfer(owner()).queueInitTransfer(
             from,
             coreNonce,
             amount128,
@@ -188,77 +160,5 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
             recipient,
             message
         );
-
-        emit PreInitTransfer(
-            id,
-            from,
-            coreNonce,
-            amount128,
-            fee,
-            recipient,
-            message
-        );
-    }
-
-    /// @notice Submits a committed HyperCore-originated transfer to OmniBridge.
-    /// Permissionless — the commitment is the only authority needed, so a stuck
-    /// transfer is never operator-gated. Arguments must hash to the commitment.
-    /// @dev Deleting before the external call is both the reentrancy and the replay
-    /// guard. If `initTransfer` reverts the delete rolls back with it, so a transient
-    /// bridge failure leaves the transfer retryable rather than burning it.
-    function triggerPendingInitTransfer(
-        uint256 id,
-        address sender,
-        uint64 coreNonce,
-        uint128 amount,
-        uint128 fee,
-        string calldata recipient,
-        string calldata message
-    ) external {
-        bytes32 committed = pendingInitTransfers[id];
-        if (committed == bytes32(0)) {
-            revert PendingInitTransferNotFound(id);
-        }
-        if (
-            committed !=
-            _initTransferCommitment(
-                sender,
-                coreNonce,
-                amount,
-                fee,
-                recipient,
-                message
-            )
-        ) {
-            revert PayloadMismatch(id);
-        }
-
-        delete pendingInitTransfers[id];
-
-        IOmniBridgeInitTransfer(owner()).initTransfer(
-            address(this),
-            amount,
-            fee,
-            0,
-            recipient,
-            message
-        );
-    }
-
-    /// @dev `abi.encode`, not `encodePacked`: two adjacent dynamic strings would let
-    /// a packed encoding collide (`"ab" + "c"` vs `"a" + "bc"`), which here would
-    /// mean submitting a different recipient than the one HyperCore committed to.
-    function _initTransferCommitment(
-        address sender,
-        uint64 coreNonce,
-        uint128 amount,
-        uint128 fee,
-        string memory recipient,
-        string memory message
-    ) private pure returns (bytes32) {
-        return
-            keccak256(
-                abi.encode(sender, coreNonce, amount, fee, recipient, message)
-            );
     }
 }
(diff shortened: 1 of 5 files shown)


evm/src/omni-bridge/contracts/HlBridgeToken.sol
```diff
@@ -39,6 +39,10 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
     uint8 public constant ACTION_TRANSFER = 0;
     uint8 public constant ACTION_INIT_TRANSFER = 1;
 
+    /// @notice coreNonce => keccak256(abi.encode(sender, coreNonce, amount, fee,
+    /// recipient, message)) of a transfer awaiting submission; zero means none.
+    mapping(uint64 => bytes32) public pendingInitTransfers;
```
kampod: I think using `coreNonce` is bad idea. Since coreNonce` is generated by user, what happens if two users randomly generate same `coreNonce`? Essentially, `coreNonce` is just timestamp.

olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/0fee227097ca1e3fa9cff91b26b5c381221cc171

evm/src/omni-bridge/contracts/HlBridgeToken.sol
```diff
@@ -121,24 +124,41 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
 
         if (action == ACTION_TRANSFER) {
             address recipient = abi.decode(tail, (address));
+            // _update to address(0) burns instead of transferring, silently
+            // dropping tokens HyperCore still counts as backed.
+            if (recipient == address(0)) revert InvalidRecipient();
```
kiseln, line 127: What happens to HyperCore transaction if we revert here?

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,163 @@
...
+/// transaction. Only that path is split; ordinary `initTransfer` is untouched.
+// slither-disable-start unused-return
+contract HlOmniBridgeWormhole is OmniBridgeWormhole {
+    /// @notice originNonce => commitment; zero means nothing pending. Enumerable
+    /// against `currentOriginNonce`.
+    mapping(uint64 => bytes32) public pendingInitTransfers;
+
+    uint256[50] private __gap;
+
+    event PreInitTransfer(
```
kiseln, line 22: I would call it consistent with function name which uses `queue`. No preference which one to rename

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,163 @@
...
+pragma solidity ^0.8.24;
+
+import {OmniBridgeWormhole} from "./OmniBridgeWormhole.sol";
+import {BridgeToken} from "./BridgeToken.sol";
+import "./BridgeTypes.sol";
+
+/// @notice OmniBridge variant for HyperEVM.
+/// @dev The HyperCore -> HyperEVM callback runs as a system transaction whose logs
+/// are absent from the block's `logsBloom`, so neither Wormhole guardians nor our
+/// indexer see anything emitted from it. HyperCore-originated transfers are
```
kiseln: our indexer can catch it from what I understand

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,163 @@
...
+    /// `coreNonce` is emitted for correlation and left out of the commitment.
+    function queueInitTransfer(
+        address sender,
+        uint64 coreNonce,
+        uint128 amount,
+        uint128 fee,
+        string calldata recipient,
+        string calldata message
+    ) external returns (uint64 originNonce) {
+        if (!isBridgeToken[msg.sender]) {
```
kiseln: Why are we not supporting HL-native tokens?

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,163 @@
...
+            revert PayloadMismatch(originNonce);
+        }
+
+        delete pendingInitTransfers[originNonce];
+
+        // The callback parked the tokens on the token contract itself.
+        BridgeToken(tokenAddress).burn(tokenAddress, amount);
+
+        initTransferExtension(
+            tokenAddress,
```
kiseln: Why are we passing tokenAddress as sender?

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,163 @@
...
+        uint128 amount,
+        uint128 fee,
+        string calldata recipient,
+        string calldata message
+    ) external returns (uint64 originNonce) {
+        if (!isBridgeToken[msg.sender]) {
+            revert NotBridgeToken(msg.sender);
+        }
+        if (fee >= amount) {
+            revert InvalidFee();
```
kiseln: consider reverting on step 2 rather than here

evm/src/omni-bridge/contracts/HlBridgeToken.sol
olga24912: The tokens will get stuck on the system address. Overall, if this case is not handled, the difference is not very significant.

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/cccefdde93d5cd2ccf3a23626b046ae1c80e24ec

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/35d942cfb72befd6c42816ca37c665a4023f0950

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/e157ba738cfea467eca09447e5de3e98ebbcb2d3

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/64373c905d7a9d01b565d4fa5a019d737229d65d

evm/src/omni-bridge/contracts/HlBridgeToken.sol
```diff
@@ -121,24 +122,41 @@ contract HyperliquedBridgeToken is BridgeToken, ICoreReceiveWithData {
...
+            _preInitTransfer(from, coreNonce, amount128, tail);
         } else {
             revert UnknownAction(action);
         }
 
         emit CoreReceived(from, action, coreNonce, amount, data);
     }
+
+    /// @dev Decodes eagerly so a malformed payload reverts while the transfer is
+    /// still atomic with the HyperCore debit. The commitment itself lives on the
```
kiseln: I think we discussed that it's not atomic, so the comment is wrong. Also it means we can fail here and have coins stuck on the system address. Should we decode on the second step instead?

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,160 @@
...
+        address indexed tokenAddress,
+        address indexed sender,
+        uint64 coreNonce,
+        uint128 amount,
+        uint128 fee,
+        string recipient,
+        string message
+    );
+
+    error NothingPending(uint64 originNonce);
```
kiseln: Not the best error name

evm/src/omni-bridge/contracts/HlBridgeToken.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/d87c983631a8afa76c3645eb0917e02e44c84810

Removed the comment. I think it's better to decode at this step, so we get a clear event. We don't have a refund path in any case, so I don't think it makes much difference at which step the funds get stuck.

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/commits/d2a87b8fa43449951350f25327f97c5916560ab4

kiseln (approved):

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
```diff
@@ -0,0 +1,186 @@
...
+/// ordinary transaction. Only that path is split; ordinary `initTransfer` is
+/// untouched.
+// slither-disable-start unused-return
+contract HlOmniBridgeWormhole is OmniBridgeWormhole {
+    /// @custom:storage-location erc7201:aurora.HlOmniBridgeWormhole
+    struct HlOmniBridgeWormholeStorage {
+        mapping(uint64 => bytes32) pendingInitTransfers;
+    }
+
+    // keccak256(abi.encode(uint256(keccak256("aurora.HlOmniBridgeWormhole")) - 1)) & ~bytes32(uint256(0xff))
```
karim-en: replace aurora with omni or near

evm/src/omni-bridge/contracts/OmniBridgeWormhole.sol
```diff
@@ -29,6 +29,10 @@ contract OmniBridgeWormhole is OmniBridge {
     uint8 private _consistencyLevel;
     uint32 public wormholeNonce;
 
+    /// @dev Reserved so that adding state here never shifts a subclass's slots.
+    /// Decrease the size by one for each variable added above it.
+    uint256[50] private __gap;
```
karim-en: remove this change

karim-en (changes requested):

evm/src/omni-bridge/contracts/HlOmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/9992a472761e225d6e8fcf78312cd7fd290a3455

evm/src/omni-bridge/contracts/OmniBridgeWormhole.sol
olga24912: fixed: https://github.com/Near-One/omni-bridge/pull/651/changes/9992a472761e225d6e8fcf78312cd7fd290a3455

karim-en (approved):

